PR #36, opened 2022-08-18
Initial commit for the refactoring of the cache setup used in tests.

denisonbarbosa:
Added some functions to read sql files and create the databases in a
temporary directory and to load the data of a dump into the created
databases.

Also fixed a little bug in a function used to read the dumps.

### DIFF
--- a/internal/cache/passwddb_test.go
+++ b/internal/cache/passwddb_test.go
@@ -49,7 +49,7 @@ func TestGetUserByName(t *testing.T) {
 			require.NoError(t, err, "GetUserByName should not have returned an error and has")
 
 			// Handle dynamic fields
-			// LastOnlineAuth should be greater than 0
+			// Checks if the lastOnlineAuth value was loaded properly.
 			assert.False(t, u.LastOnlineAuth.IsZero(), "Last Online should not be zero.")
 			u.LastOnlineAuth = time.Unix(0, 0)
 
@@ -112,7 +112,7 @@ func TestGetUserByUID(t *testing.T) {
 			require.NoError(t, err, "GetUserByName should not have returned an error and has")
 
 			// Handle dynamic fields
-			// LastOnlineAuth should be greater than 0
+			// Checks if the lastOnlineAuth value was loaded properly.
 			assert.False(t, u.LastOnlineAuth.IsZero(), "Last Online should not be zero.")
 			u.LastOnlineAuth = time.Unix(0, 0)
 
@@ -169,7 +169,7 @@ func TestNextPasswdEntry(t *testing.T) {
 		u, err := c.NextPasswdEntry(context.Background())
 		require.NoError(t, err, "NextPasswdEntry should initiate and returns values without any error")
 
-		// LastOnlineAuth should be greater than 0
+		// Checks if the lastOnlineAuth value was loaded properly.
 		assert.False(t, u.LastOnlineAuth.IsZero(), "Last Online should not be zero.")
 		u.LastOnlineAuth = time.Unix(0, 0)
 
--- a/internal/testutils/db.go
+++ b/internal/testutils/db.go
@@ -211,14 +211,12 @@ func PrepareDBsForTests(t *testing.T, cacheDir, initialCache string, options ...
 	_, p, _, _ := runtime.Caller(0)
 	testutilsPath := filepath.Dir(p)
 
-	options = append(options, cache.WithTeardownDuration(0))
-
 	c := NewCacheForTests(t, cacheDir, options...)
 	err := c.Close(context.Background())
 	require.NoError(t, err, "Cache must be closed to enable the dump loading.")
 
 	for _, db := range []string{"passwd.db", "shadow.db"} {
-		LoadDumpIntoDB(t, filepath.Join(testutilsPath, "cache_dumps", initialCache, db+".dump"), filepath.Join(cacheDir, db))
+		loadDumpIntoDB(t, filepath.Join(testutilsPath, "cache_dumps", initialCache, db+".dump"), filepath.Join(cacheDir, db))
 	}
 }
 
@@ -228,7 +226,7 @@ func NewCacheForTests(t *testing.T, cacheDir string, options ...cache.Option) (c
 
 	uid, gid := GetCurrentUIDGID(t)
 	opts := append([]cache.Option{}, cache.WithCacheDir(cacheDir),
-		cache.WithRootUID(uid), cache.WithRootGID(gid), cache.WithShadowGID(gid))
+		cache.WithRootUID(uid), cache.WithRootGID(gid), cache.WithShadowGID(gid), cache.WithTeardownDuration(0))
 
 	opts = append(opts, options...)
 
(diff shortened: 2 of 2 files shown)


nss/testdata/users_in_db/shadow.db.dump
```diff
@@ -0,0 +1,6 @@
+shadow
```
didrocks, line 1: Our goal is now to have a single dump reference, in testutils/ that we use.

didrocks (changes requested):

internal/cache/shadowdb_test.go
```diff
@@ -127,9 +128,9 @@ func TestNextShadowCloseBeforeIterationEnds(t *testing.T) {
 	t.Parallel()
 
 	cacheDir := t.TempDir()
-	insertUsersInDb(t, cacheDir)
+	testutils.PrepareDBsForTests(t, cacheDir, "users_in_db")
 
-	c := newCacheForTests(t, cacheDir, cache.WithTeardownDuration(0), cache.WithOfflineCredentialsExpiration(0))
+	c := testutils.NewCacheForTests(t, cacheDir)
```
didrocks, line 133: This was called with the 2 options which are not in `NewCacheForTests()`, I know we don’t need `cache.WithOfflineCredentialsExpiration` anymore with "user_in_db", but the teardown duration to 0 is better.

internal/cache/shadowdb_test.go
```diff
@@ -118,7 +119,7 @@ func TestNextShadowEntry(t *testing.T) {
 func TestNextShadowEntryNoShadow(t *testing.T) {
 	t.Parallel()
 
-	c := newCacheForTests(t, t.TempDir(), cache.WithTeardownDuration(0), cache.WithOfflineCredentialsExpiration(0))
+	c := testutils.NewCacheForTests(t, t.TempDir())
```
didrocks, line 122: This was called with the 2 options which are not in `NewCacheForTests()`, I know we don’t need here as we don’t have any entry, but the teardown duration to 0 is better.

internal/cache/shadowdb_test.go
```diff
@@ -89,9 +90,9 @@ func TestNextShadowEntry(t *testing.T) {
 	}
 
 	cacheDir := t.TempDir()
-	insertUsersInDb(t, cacheDir)
+	testutils.PrepareDBsForTests(t, cacheDir, "users_in_db")
 
-	c := newCacheForTests(t, cacheDir, cache.WithTeardownDuration(0), cache.WithOfflineCredentialsExpiration(0))
+	c := testutils.NewCacheForTests(t, cacheDir)
```
didrocks, line 95: This was called with the 2 options which are not in `NewCacheForTests()`, I know we don’t need `cache.WithOfflineCredentialsExpiration` anymore with "user_in_db", but the teardown duration to 0 is better.

internal/cache/shadowdb_test.go
```diff
@@ -33,9 +34,9 @@ func TestGetShadowByName(t *testing.T) {
 			t.Parallel()
 
 			cacheDir := t.TempDir()
-			insertUsersInDb(t, cacheDir)
+			testutils.PrepareDBsForTests(t, cacheDir, "users_in_db")
 
-			c := newCacheForTests(t, cacheDir, cache.WithTeardownDuration(0), cache.WithOfflineCredentialsExpiration(0))
+			c := testutils.NewCacheForTests(t, cacheDir)
```
didrocks, line 39: This was called with the 2 options which are not in `NewCacheForTests()`, I know we don’t need `cache.WithOfflineCredentialsExpiration` anymore with "user_in_db", but the teardown duration to 0 is better.

internal/cache/passwddb_test.go
```diff
@@ -163,20 +159,18 @@ func TestNextPasswdEntry(t *testing.T) {
...
 	// Iterate over all entries
 	numIteration := len(wanted)
 	for i := 0; i < numIteration; i++ {
 		u, err := c.NextPasswdEntry(context.Background())
 		require.NoError(t, err, "NextPasswdEntry should initiate and returns values without any error")
 
-		// LastOnlineAuth should be recent
-		assert.True(t, testutils.TimeBetweenOrEquals(u.LastOnlineAuth, startTime, endTime), "Last Online auth should match insertion time. Last Online auth: %v. Start: %v, End: %v", u.LastOnlineAuth, startTime, endTime)
+		// LastOnlineAuth should be greater than 0
+		assert.False(t, u.LastOnlineAuth.IsZero(), "Last Online should not be zero.")
```
didrocks, line 173: oh, there is no Greater/Less?

internal/testutils/db.go
```diff
@@ -199,3 +202,79 @@ func predicatableFieldValues(name string, data, cols []string) {
...
+	}
+}
+
+// NewCacheForTests returns a cache that is closed automatically, with permissions set to current user.
+func NewCacheForTests(t *testing.T, cacheDir string, options ...cache.Option) (c *cache.Cache) {
+	t.Helper()
+
+	uid, gid := GetCurrentUIDGID(t)
+	opts := append([]cache.Option{}, cache.WithCacheDir(cacheDir),
+		cache.WithRootUID(uid), cache.WithRootGID(gid), cache.WithShadowGID(gid))
```
didrocks, line 231: Actually, we can set WithTearDownDuration to 0 here, and then, only in tests when it’s not 0, we can append another value to it, which should suppress previous options setting. Wdyt?

That way, the default settings for the test function makes sense.

internal/testutils/db.go
```diff
@@ -130,7 +132,8 @@ func dbToCsv(t *testing.T, db *sql.DB, w io.Writer, usePredicatableFieldValues b
 		err = dumpTable(t, db, tableName, w, usePredicatableFieldValues)
 		require.NoError(t, err, "Failed to dump table %s", tableName)
 
-		separateTables = true
+		_, err = w.Write([]byte("\n"))
```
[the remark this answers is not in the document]
denisonbarbosa: Ok. I thought it would be better to put it in a different PR, but I needed those changes in order to make the other functions work, but I should have opened another PR and requested it to be merged before this one.

internal/testutils/db.go
```diff
@@ -199,3 +202,79 @@ func predicatableFieldValues(name string, data, cols []string) {
...
+
+	c, err := cache.New(context.Background(), opts...)
+	require.NoError(t, err, "Setup: should be able to create a cache")
+	t.Cleanup(func() { c.Close(context.Background()) })
+
+	return c
+}
+
+// LoadDumpIntoDB reads the specified dump file and inserts its contents into the database.
+func LoadDumpIntoDB(t *testing.T, dumpPath, dbPath string) {
```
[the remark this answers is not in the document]
denisonbarbosa: Sure. I was using it in other packages before, but after the refactoring it does not need to be public anymore. It was my bad, I let it slip while I was reviewing the code before pushing.

internal/cache/shadowdb_test.go
denisonbarbosa: Yeah. My intention was to leave it as a default to the cache for tests, but I've added the option in the wrong function 🥲

internal/cache/passwddb_test.go
denisonbarbosa: There is, but then I would have to either compare it with the 0 value for time, which is weird, or I would have to convert the time to Unix and them compare it with 0. Since the time package already has this builtin function (which, I think, would compare the Time more efficiently), I thought it would be better to use it.

internal/testutils/db.go
denisonbarbosa: That was the intention... but I've ended up adding the option in the wrong function (my bad).

internal/cache/passwddb_test.go
didrocks: Looks at the rest of the code to get inspiration :)

Right now, I’m comparing to the 0 time, but I do not have strong preference. However, we should only do it one way throughout the code.

didrocks (approved):
Perfect, every comments are now addressed, thanks for the changed!

PR #37, opened 2022-08-18
Implement aad-auth command line tool

GabrielNagy:
CLI commands:
```sh-session
> aad-cli user [command]
Available Commands:
  get         Query local Azure AD user settings
  set         Configure local Azure AD user settings (can only set home, shell and gecos for now)
              Only interacts with the passwd database.

> aad-cli config [command]
Available Commands:
  edit        Edit the configuration file in an editor
  print       Print the current configuration, parsed from /etc/aad.conf

> aad-cli version
Returns version of aad-cli and exits (will also check for pam/nss module versions if installed via dpkg/apt)
```

Autocompletion is implemented on the `user get` and `user set` commands. Execute the following in the aad-cli directory to test for bash (replace with fish/zsh if needed):
```sh
export PATH=.:$PATH
. <(aad-cli completion bash)
```

### DIFF
--- a/cmd/aad-cli/main.go
+++ b/cmd/aad-cli/main.go
@@ -12,6 +12,9 @@ import (
 	"github.com/ubuntu/aad-auth/internal/logger"
 )
 
+//go:generate go run ../generate_completion_documentation.go completion ../../generated
+//go:generate go run ../generate_completion_documentation.go man ../../generated
+
 func run(a *cli.App) int {
 	defer installSignalHandler(a)()
 
--- a/debian/aad-cli.install
+++ b/debian/aad-cli.install
@@ -1,0 +1,5 @@
+usr/bin/aad-cli
+usr/share/zsh/vendor-completions/_aad-cli
+usr/share/bash-completion/completions/aad-cli
+usr/share/fish/vendor_completions.d/aad-cli.fish
+usr/share/man/man1/aad-cli.1
--- a/debian/control
+++ b/debian/control
@@ -43,10 +43,22 @@ Description: ${source:Synopsis} module for NSS
 
 Package: aad-common
 Architecture: all
-Enhances: libpam-aad, libnss-aad
+Enhances: libpam-aad, libnss-aad, aad-cli
 Depends: ${misc:Depends}
 Description: ${source:Synopsis} common files
  ${source:Extended-Description}
  .
  This package contains configuration and translation files shared between the
  other AAD packages.
+
+Package: aad-cli
+Architecture: any
+Built-Using: ${misc:Built-Using},
+Depends: aad-common,
+         ${shlibs:Depends},
+         ${misc:Depends},
+Description: ${source:Synopsis} CLI tool
+ ${source:Extended-Description}
+ .
+ This package contains the CLI tool to manage various aspects of the Azure AD
+ integration.
--- a/debian/rules
+++ b/debian/rules
@@ -2,7 +2,7 @@
 #export DH_VERBOSE = 1
 
 export GOCACHE := $(shell mktemp -d /tmp/gocache-XXXX)
-export GOFLAGS := -buildmode=pie
+export GOFLAGS := -ldflags=-X=github.com/ubuntu/aad-auth/internal/consts.Version=$(shell dpkg-parsechangelog -S Version) --mod=vendor -buildmode=pie
 
 export DEB_BUILD_MAINT_OPTIONS := optimize=-lto
 
@@ -31,7 +31,7 @@ override_dh_auto_build:
 	make -C nss/
 
 	# Build the CLI
-	# DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/cmd/aad-auth dh_auto_build
+	DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/cmd/aad-cli dh_auto_build
 
 	# Build the NSS helper binary
 	DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/nss/aad-auth dh_auto_build
@@ -56,5 +56,5 @@ override_dh_auto_install:
 	mkdir -p debian/tmp/usr/libexec
 	mv debian/tmp/usr/bin/aad-auth debian/tmp/usr/libexec
 
-	# Generate and install translations
-	GENERATE_ONLY_INSTALL_TO_DESTDIR=$(CURDIR)/debian/tmp go generate -x ./internal/i18n
+	# Generate and install translations and shell completions
+	GENERATE_ONLY_INSTALL_TO_DESTDIR=$(CURDIR)/debian/tmp go generate -x ./internal/i18n ./cmd/aad-cli


cmd/aad-cli/cli/config.go
```diff
@@ -0,0 +1,160 @@
...
+		Use:   "edit",
+		Short: "Edit the configuration file in an editor",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// Create a temporary file with the previous config file contents
+			tempfile, err := tempFileWithPreviousConfig(a.options.configFile)
+			if err != nil {
+				return err
+			}
+
```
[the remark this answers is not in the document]
didrocks: ahhhhhhh, I understand now why it’s later on. So yeah, let’s not remove it if something fails. We will have the finale rename anyway. as you say.

cmd/aad-cli/cli/config.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+			// TODO see if it's worth checking if the files are the same before doing a pointless write
+			defer os.Remove(tempfile)
+
+			newConfig, err := os.ReadFile(tempfile)
+			if err != nil {
+				return fmt.Errorf("Error: failed to read temporary config file: %w", err)
+			}
+			//#nosec:G306 these are the expected permissions for the config file
+			if err := os.WriteFile(a.options.configFile, newConfig, 0640); err != nil {
```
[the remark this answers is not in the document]
didrocks: Actually, you need both files to be on the same filesystems, as you say. The practice is generally to create the temporary file in the same directory. This one can have a fix named (like with a suffix). This is what `vim`, `dpkg` and others are doing. So, we should do the same.

cmd/aad-cli/cli/config.go
```diff
@@ -0,0 +1,160 @@
...
+	cfg, err := config.ToIni()
+	if err != nil {
+		return err
+	}
+
+	if _, err := cfg.WriteTo(buf); err != nil {
+		return fmt.Errorf("could not write config to buffer: %w", err)
+	}
+	fmt.Println("[" + domainSection + "]")
+	fmt.Print(buf.String())
```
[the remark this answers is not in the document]
didrocks: Yeah, but that’s one of the only benefit for the print command over using `cat` and looking at the correct section. There is only some code in `config/` which does that, can we reuse the output of it?

conf/aad.conf.template
```diff
@@ -1,16 +1,16 @@
...
-offline_credentials_expiration = <expiration interval for someDomain.com>
-homedir = /home/someDomain/%u
-shell = /bin/ssh
+## overriding values for a specific domain, every value inside a section is optional
+# [someDomain.com]
+# tenant_id = <TENANT_ID for someDomain.com>
+# app_id = <APP OR CLIENT ID for someDomain.com>
+# offline_credentials_expiration = <expiration interval for someDomain.com>
+# homedir = /home/someDomain/%u
+# shell = /bin/ssh
```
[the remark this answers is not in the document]
didrocks: Sounds good, can add comments on the main section for each option? Something like `/etc/logrotate.conf` for instance.

Btw, you can replace someDomain with `%somethingIdontRemeberLookAtTheCode` maybe? :p

I suggest that tenant-id and app-id are of form xxxxx-yyy-... with the real format. Wdyt?

internal/cache/cache_test.go
```diff
@@ -395,3 +395,124 @@ func TestCanAuthenticate(t *testing.T) {
...
+				return
+			}
+			assert.NoError(t, err, "UpdateUserAttribute should not have returned an error but has")
+
+			got, err := c.GetUserByName(context.Background(), tc.username)
+			require.NoError(t, err, "Setup: GetUserByName should not have returned an error but has")
+
+			switch tc.attribute {
+			case "gecos":
+				require.Equal(t, "new gecos", got.Gecos)
```
didrocks, line 443: you can replace with tc.value, directly.

internal/config/config_test.go
```diff
@@ -150,6 +151,50 @@ func TestLoadConfig(t *testing.T) {
 	}
 }
 
+func TestToIni(t *testing.T) {
```
didrocks, line 154: could be parallel

internal/config/config_test.go
```diff
@@ -150,6 +151,50 @@ func TestLoadConfig(t *testing.T) {
...
+
+func TestValidate(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		configFile string
+		wantErr    bool
+	}{
+		"valid config, default domain":   {configFile: "valid.conf"},
+		"valid config, multiple domains": {configFile: "valid-multiple-domains.conf"},
```
didrocks, line 176: Thinking about it, do you check that there are required fields uncommented? I think Validate() should do this.

internal/config/config_test.go
```diff
@@ -150,6 +151,50 @@ func TestLoadConfig(t *testing.T) {
...
+
+		// Error cases
```
didrocks, line 178: ditto on home parsing: we should ensure there is only `%valuewesupport`

cmd/aad-cli/cli/config_test.go
```diff
@@ -136,15 +134,17 @@ func TestConfigEditor(t *testing.T) {
...
-	t.Cleanup(func() { os.Remove(editor.Name()) })
 
 	var b bytes.Buffer
-	b.WriteString("#!/bin/sh\n")
+	b.WriteString("#!/bin/sh")
 
-	b.WriteString(`echo "TEMPORARY CONFIG PATH: $1"`)
+	b.WriteString(`
+echo "TEMPORARY CONFIG PATH: $1"
+echo "TEMPORARY CONFIG CONTENTS:"
```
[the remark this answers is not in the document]
didrocks: ack then :)

internal/config/config_test.go
didrocks: Indeed, mind adding a test for it? (there is one when loading the config, but I guess it’s a great illustration of its usage)

internal/config/config_test.go
didrocks: Agreed.

internal/cache/cache_test.go
```diff
@@ -395,3 +395,124 @@ func TestCanAuthenticate(t *testing.T) {
...
+				return
+			}
+			assert.NoError(t, err, "QueryPasswdAttribute should not have returned an error but has")
+
+			got, err := c.GetUserByName(context.Background(), tc.username)
+			require.NoError(t, err, "Setup: GetUserByName should not have returned an error but has")
+
+			switch tc.attribute {
+			case "login":
+				require.Equal(t, value, got.Name)
```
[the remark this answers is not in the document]
didrocks: I think for the date, timestamp is the easiest way to deserialize…

cmd/aad-cli/cli/user.go
```diff
@@ -0,0 +1,144 @@
...
+			if err != nil {
+				return err
+			}
+
+			fmt.Println(strings.TrimSpace(fmt.Sprint(value)))
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&a.username, "name", "n", getDefaultUser(), "username to operate on")
+	cmd.Flags().BoolVarP(&a.allusers, "all", "a", false, "list all users")
```
didrocks, line 97: for both of them, same remark than in the other files: keep that as local variable in `install()` function.

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/ubuntu/aad-auth/cmd/aad-cli/cli"
+	"github.com/ubuntu/aad-auth/internal/cache"
+	"github.com/ubuntu/aad-auth/internal/testutils"
+)
+
+func TestUserShellCompletion(t *testing.T) {
+	tests := map[string]struct {
+		args []string
+	}{
+		"get all users for get, short flag":    {args: []string{"user", "-n"}},
```
didrocks, line 22: it’s not "for get" anymore

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+			cacheDir := t.TempDir()
+			testutils.CopyDBAndFixPermissions(t, filepath.Join("testdata", "cachedb"), cacheDir)
+
+			uid, gid := testutils.GetCurrentUIDGID(t)
+
+			cache, err := cache.New(context.Background(), cache.WithCacheDir(cacheDir), cache.WithRootUID(uid), cache.WithRootGID(gid), cache.WithShadowGID(gid))
+			require.NoError(t, err, "Setup: failed to create cache")
+
+			c := cli.New(cli.WithCache(cache))
```
didrocks, line 44: Ah, I understand now why you are attaching cache to the app itself. Ack on it then. (still, my comments to remove option is valid). I kept my previous remark to see that I was first puzzled by it and that maybe a comment will be beneficial?

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			args := []string{cobra.ShellCompRequestCmd}
+			args = append(args, tc.args...)
+			args = append(args, "")
+
+			cacheDir := t.TempDir()
+			testutils.CopyDBAndFixPermissions(t, filepath.Join("testdata", "cachedb"), cacheDir)
```
didrocks, line 37: I’m sorry, but this will need to be rebased now that we merged Denison's work :) (more on that in multiple parts of the PR)

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+			uid, gid := testutils.GetCurrentUIDGID(t)
+
+			cache, err := cache.New(context.Background(), cache.WithCacheDir(cacheDir), cache.WithRootUID(uid), cache.WithRootGID(gid), cache.WithShadowGID(gid))
```
didrocks, line 41: ditto on how we load the cache.

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+			got, err := testutils.RunApp(t, c, args...)
+			require.NoError(t, err, "failed to run completion")
+
+			want := testutils.SaveAndLoadFromGolden(t, got)
+			require.Equal(t, want, got, "expected output to match golden file")
+		})
+	}
+}
+
+func TestUserGet(t *testing.T) {
```
didrocks, line 54: the function can be rename (Simply "TestUser"?)

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+				got = timestampToUnix(t, got, user.LastOnlineAuth)
+			}
+
+			want := testutils.SaveAndLoadFromGolden(t, got)
+			require.Equal(t, want, got, "expected output to match golden file")
+		})
+	}
+}
+
+func TestUserSet(t *testing.T) {
```
didrocks, line 131: and here, I would say "TestUserSetAttribute" (even if that doesn’t match fully the command line, if you feel both tests should be kept separated as not having enough in common, this is fine)

cmd/aad-cli/cli/testdata/golden/TestUserGet/get_gecos
```diff
@@ -0,0 +1 @@
+
```
didrocks, line 1: hem, it’s empty in the test case, it probably should be set to compare it to something valuable :)

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+	tests := map[string]struct {
+		username           string
+		attribute          string
+		shadowNotAvailable bool
+		allUsers           bool
+
+		wantErr bool
+	}{
+		"get all users":                  {allUsers: true},
+		"get user":                       {username: "myuser@domain.com"},
```
didrocks, line 64: I think rather than splitting in username/attribute, having the whole command line here []string or just string is easier when you read the test for what accepts the command line, a little bit your previous example. what do you think?

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+func TestUserSet(t *testing.T) {
+	tests := map[string]struct {
+		username  string
+		attribute string
+
+		badPerms bool
+		wantErr  bool
+	}{
+		"set gecos": {attribute: "gecos"},
```
didrocks, line 139: same remark about username/attribute than in the previous test

cmd/aad-cli/cli/testdata/golden/TestUserShellCompletion/get_attributes_for_overridden_user
```diff
@@ -0,0 +1,9 @@
+login
+password
+uid
+gid
+gecos
+home
+shell
+last_online_auth
+:4
```
didrocks, line 9: I have one very stupid question: where do you set the return code in stdout? I don’t find it in RunApp() testutils helper, nor in the cmd.Run() command, so I’m puzzled :)

didrocks (changes requested):
Excellent work! I see that all patterns we discussed in the previous months are completely applied here and I have not much to say (despite the number of comments) for the first real review on this big PR. Thanks a lot for the level of care and attention to details!

In addition to the comments/praises and questions in the PR, you will then need to rebase on top of https://github.com/ubuntu/aad-auth/pull/36#pullrequestreview-1088724978 now that it’s merged, which will slightely change the way you load the cache and dump existing db.

Excellent work again :)

internal/cache/cache.go
```diff
@@ -71,7 +71,12 @@ type Option func(*options) error
 // WithCacheDir specifies a personalized cache directory.
 func WithCacheDir(p string) func(o *options) error {
 	return func(o *options) error {
-		o.cacheDir = p
+		cacheDir := p
```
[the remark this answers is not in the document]
didrocks: ack :)

cmd/aad-cli/cli/version_test.go
```diff
@@ -0,0 +1,80 @@
...
+	require.NoError(t, err, "Setup: failed to write temporary file")
+
+	return tmpfile.Name()
+}
+
+func sanitizeDevVersion(s string) string {
+	return strings.ReplaceAll(s, consts.Version, "dev")
+}
+
+func TestMain(m *testing.M) {
```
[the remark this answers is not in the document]
didrocks: hum, let me guess let me guess let me guess… Help? :p

cmd/aad-cli/cli/config.go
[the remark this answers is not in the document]
didrocks: So, let’s keep this as a future bonus :)

.gitignore
```diff
@@ -7,9 +7,11 @@
 *.so.*
 *.dylib
 
+# Editor exclusions
+.vsls.json
+
 # Test binary, built with `go test -c`
 *.test
-coverage/
```
didrocks, line 12: oh, you removed coverage/ created by gotest? Or was it doubled?

cmd/aad-cli/cli/config.go
```diff
@@ -0,0 +1,144 @@
...
+	cmd.MarkFlagsMutuallyExclusive("edit", "domain")
+	a.rootCmd.AddCommand(cmd)
+}
+
+// tempFileWithPreviousConfig returns a temporary file with the contents of the
+// previous config file if it exists.
+// If the previous config file does not exist, its contents are empty.
+// If the previous config file cannot be read, an error is returned.
+func tempFileWithPreviousConfig(configFile string) (string, error) {
+	tempfile, err := os.CreateTemp(filepath.Dir(configFile), "aad.*.conf")
```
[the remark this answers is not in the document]
didrocks: sounds good to me!

cmd/aad-cli/cli/testdata/golden/TestConfigPrint/some_optional_fields_missing
```diff
@@ -0,0 +1,6 @@
+[default]
```
didrocks, line 1: Stupid question, but what field is missing?

internal/config/config.go
```diff
@@ -74,28 +83,9 @@ func Load(ctx context.Context, p, domain string, opts ...Option) (config AAD, er
...
-			} else {
-				config.OfflineCredentialsExpiration = &v
-			}
-		}
-		if tmp := cfgSection.Key("homedir").String(); tmp != "" {
-			config.HomeDirPattern = tmp
-		}
-		if tmp := cfgSection.Key("shell").String(); tmp != "" {
-			config.Shell = tmp
+	for _, section := range []string{ini.DefaultSection, domain} {
```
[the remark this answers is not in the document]
didrocks: No no, this is perfect, let’s keep it that way :)

internal/config/config_test.go
```diff
@@ -142,14 +145,62 @@ func TestLoadConfig(t *testing.T) {
...
+
+func TestValidate(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		configFile string
+		wantErr    bool
+	}{
+		"valid config, default domain":   {configFile: "valid.conf"},
+		"valid config, multiple domains": {configFile: "valid-multiple-domains.conf"},
```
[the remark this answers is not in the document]
didrocks: Yeah, this looks good :) Thanks for testing!

cmd/aad-cli/cli/testdata/golden/TestConfigPrint/some_optional_fields_missing
didrocks: Thanks! :)

didrocks: Actually, can you add them commented so that we know in the future?

didrocks: And I don’t know if the diff is correct, but I see a:
```
homedir                        = /home/%f
shell                          = /bin/bash
```

in the diff?

cmd/aad-cli/cli/user.go
```diff
@@ -0,0 +1,220 @@
...
+		// Update the home directory if it changed
+		if err := moveUserHome(ctx, username, fmt.Sprintf("%v", prevValue), fmt.Sprintf("%v", value)); err != nil {
+			return fmt.Errorf("Unable to move home directory for %s: %w", username, err)
+		}
+	}
+	return nil
+}
+
+// moveUserHome moves the home directory of an user from the previous location to the new one.
+func moveUserHome(ctx context.Context, username, prevValue, value string) error {
```
didrocks, line 192: I would add a defer func() {} here instead for `fmt.Errorf("Unable to move home directory for %s: %w", username, err) ` and just return the err directly in the function call.

This way, we are in line with the function itself decorating itself (if reused one day elsewhere)

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+func TestUserSetAttribute(t *testing.T) {
+	tests := map[string]struct {
+		args string
+
+		badPerms bool
+		wantErr  bool
+	}{
```
didrocks, line 124: So, the only can we don’t test this way is for the current user detection. Any idea how we could do this?

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func TestUser(t *testing.T) {
+	tests := map[string]struct {
+		args               string
+		shadowNotAvailable bool
+
+		wantErr bool
+	}{
+		"get all users":                  {args: "--all"},
```
didrocks, line 58: So, the only can we don’t test this way is for the current user detection. Any idea how we could do this (as it’s not in db)?

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+		// Error cases
+		"fail if previous directory is absent": {prevHomeDir: "absent", newHomeDir: "newhome", wantErr: true},
+		"fail if previous directory is a file": {prevHomeDir: "oldhomefile", newHomeDir: "newhome", wantErr: true},
+		"fail if new directory already exists": {prevHomeDir: "oldhome", newHomeDir: "existingnewhome", wantErr: true},
+	}
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			cacheDir := t.TempDir()
```
didrocks, line 181: Mind renaming this tmpDir and having the cache in a subdirectory. I find it weird to create the home in cache :)

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,275 @@
...
+			prevHomeDir := filepath.Join(cacheDir, tc.prevHomeDir)
+			newHomeDir := filepath.Join(cacheDir, tc.newHomeDir)
+
+			err = cache.UpdateUserAttribute(context.Background(), "futureuser@domain.com", "home", prevHomeDir)
+			require.NoError(t, err, "Setup: failed to set initial user home directory")
+
+			c := cli.New(cli.WithCache(cache))
+			_, runErr := testutils.RunApp(t, c, "user", "--name", "futureuser@domain.com", "home", newHomeDir, "--move-home")
+
+			// We always expect the passwd attribute to be updated in this test, regardless of moving errors.
```
didrocks, line 203: Is it how usermod works too? The reference is changed before the actual action?

cmd/aad-cli/cli/user_test.go
```diff
@@ -0,0 +1,275 @@
...
+			home, err := cache.QueryPasswdAttribute(context.Background(), "futureuser@domain.com", "home")
+			require.NoError(t, err, "Setup: failed to get user home directory")
+			require.Equal(t, newHomeDir, home, "expected home directory to be updated")
+
+			if !tc.wantErr {
```
didrocks, line 208: why did you reverse the logic? I was puzzled at first :)

cmd/aad-cli/cli/testdata/golden/TestConfigPrint/some_optional_fields_missing
didrocks: oh, I am in the golden, silly me :)

didrocks:
last thing: conflict in go.mod and go.sum since we merged the new msal in case you haven’t noticed. And done with the review, excellent work, I only have new questions with the new changes, all the previous comments are addressed. Thanks a lot!

cmd/aad-cli/cli/user_test.go
didrocks: Thanks for checking!

cmd/aad-cli/cli/testdata/golden/TestConfigPrint/some_optional_fields_missing
didrocks: sounds like a good idea! (I would unresolve to keep track then)

cmd/aad-cli/cli/user_test.go
didrocks: I meant "on the tests", like the detection of the current user. But it seems overkill TBH, wdyt?

cmd/aad-cli/cli/user_test.go
```diff
@@ -174,6 +175,7 @@ func TestUserMoveHomeDirectory(t *testing.T) {
 		"fail if previous directory is absent": {prevHomeDir: "absent", newHomeDir: "newhome", wantErr: true},
 		"fail if previous directory is a file": {prevHomeDir: "oldhomefile", newHomeDir: "newhome", wantErr: true},
 		"fail if new directory already exists": {prevHomeDir: "oldhome", newHomeDir: "existingnewhome", wantErr: true},
+		"fail if the user has open processses": {prevHomeDir: "oldhome", newHomeDir: "newhome", userLoggedIn: true, wantErr: true},
```
didrocks, line 178: Nitpick (just to have one): I would add something here too that we keep the old attribute. (hence the userLoggedIn: true, but I find it a little bit puzzling)

didrocks:
Review done! Excellent work on the changes, the additional tests in ´IsBusy()´ and such :)

There is still 2 comments to discuss about testing the current user detection if possible (unsure) and there is on nitpick :p Otherwise, this is perfect, congrats!

cmd/aad-cli/cli/user.go
```diff
@@ -104,7 +105,7 @@ func (a *App) getCache() (*cache.Cache, error) {
 
 // getDefaultUser returns the current user name or a blank string if an error occurs.
 func getDefaultUser() string {
-	u, err := user.Current()
+	u, err := osuser.Current()
```
didrocks, line 108: Someone shadowed their local "user" variable and was puzzled by it, I guess :p

cmd/aad-cli/cli/user.go
```diff
@@ -61,9 +61,9 @@ Currently the only modifiable attributes are: %s.`, strings.Join(cache.PasswdUpd
 			return runUser(a.ctx, args, c, a.options.procFs, username, allUsers, moveHome)
 		},
 	}
-	cmd.Flags().StringP("name", "n", getDefaultUser(), "username to operate on")
+	cmd.Flags().StringP("name", "n", a.options.currentUser, "username to operate on")
 	cmd.Flags().BoolP("all", "a", false, "list all users")
-	cmd.Flags().BoolP("move-home", "m", false, "if updating hoem, move the content of the home directory to the new location")
+	cmd.Flags().BoolP("move-home", "m", false, "if updating home, move the content of the home directory to the new location")
```
didrocks, line 66: lalalalala :)

cmd/aad-cli/cli/user_test.go
```diff
@@ -134,13 +135,20 @@ func TestUserSetAttribute(t *testing.T) {
...
+			if usernameIndex == -1 {
+				username = "futureuser@domain.com"
+			}
 
 			cacheDir := t.TempDir()
 			cacheDB := "db_with_old_users"
 			testutils.PrepareDBsForTests(t, cacheDir, cacheDB)
 			cache := testutils.NewCacheForTests(t, cacheDir)
-			c := cli.New(cli.WithCache(cache))
+			c := cli.New(cli.WithCache(cache), cli.WithCurrentUser("futureuser@domain.com"))
```
didrocks, line 151: Perfect! This is the way to go to allow this use case (there is unfortunately no way in the integration tests to change what `user.Current()` returns, as it’s a syscal. Yes, not the first time we ecounter this!

didrocks (approved):
Looks good! Everything is now resolved. I have no strong opinion on the error display, but I think what you do make sense (I wonder why we did it differently in adsys, that worth a check I guess, but this could come from the days where there was no mutual exclusion of flags for instance, and we wanted to point it that was the issue), do you mind giving it a final try? (unknown parameter, mutually exclusive flags and too many arguments for instance) and check that the error is obviously printed, in addition to the usage by the new Cobra.

If so, backporting this to adsys CLIs could be interesting.

Thanks again for the excellent work! As said, let’s not merge it right away anyway as we need to cut a release before asking for the FFe on this one.

didrocks (commented):
Last commit is good to me.

I understand why you are specifically picking the full path for completion files and man pages (as we might have other binaries in other packages…). Thanks! And well done again :)

PR #41, opened 2022-08-24
Refactoring dump read functions to use io.Reader rather than paths

denisonbarbosa:
Also adjusted integration_test.go to properly interact with the changes.

### DIFF
--- a/pam/integration_test.go
+++ b/pam/integration_test.go
@@ -138,16 +138,16 @@ func TestPamSmAuthenticate(t *testing.T) {
 				ref := filepath.Join(cacheDir, db)
 				wants[db] = testutils.LoadAndUpdateFromGoldenDump(t, ref)
 
-				// TODO: make this better, using a reader
 				// Load temporary got to memory
-				tmp := filepath.Join(t.TempDir(), db+".dump")
-				f, err := os.Create(tmp)
+				f, err := os.Create(filepath.Join(t.TempDir(), db+".dump"))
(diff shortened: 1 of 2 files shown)


pam/integration_test.go
```diff
@@ -138,16 +138,16 @@ func TestPamSmAuthenticate(t *testing.T) {
 				ref := filepath.Join(cacheDir, db)
 				wants[db] = testutils.LoadAndUpdateFromGoldenDump(t, ref)
 
-				// TODO: make this better, using a reader
 				// Load temporary got to memory
-				tmp := filepath.Join(t.TempDir(), db+".dump")
-				f, err := os.Create(tmp)
+				f, err := os.Create(filepath.Join(t.TempDir(), db+".dump"))
```
didrocks, line 142: didn’t we say we would use an in memory buffer which can write an read from?

didrocks (changes requested):
changes in `db.go` are perfect.

However, in `integration_test.go`, the whole goal is to use a in memory buffer rather than a file, mind changing it to this?

pam/integration_test.go
denisonbarbosa: Yeah, we did. Probably was a brain fry moment. My bad... 
(In my defense, we did say a lot of stuff yesterday 🥲).

didrocks (approved):
Thanks for the changes! Let’s merge it :)

PR #43, opened 2022-08-24
Parse string as int64 instead of int

GabrielNagy:
`strconv.Atoi` returns an int, which on 32-bit systems maps to int32, which is not large enough to hold the last_online_auth value from the DB dumps.

To avoid this issue, switch to using `strconv.ParseInt` and pass an appropriate bit size.


didrocks (approved):
Thanks for looking at the armhf failure !
This is totally my fault as I made Denison changes this. But yeah, the joy of 32 vs 64 bits. The build now pass. Approving!

PR #48, opened 2022-08-26
Fix segfault on interrupted pam conversation

jibel:
When the conversation is interrupt, the password is null and make the
module crashes.

Co-authored-by: Didier Roche <didrocks@ubuntu.com>


didrocks:
Test pass, we haven’t completely gone crazy :)

PR #50, opened 2022-08-29
Support gettext translations

### LINKED ISSUE #29: Support translations
Add support for translations:

- Generate PO files
- Integrate to CI with auto updates (cf adsys)
- Bind to text domain 
- Verify that user messages are translated.

GabrielNagy:
At a high level, this PR accomplishes the following:
- create a single textdomain for use with _all_ packages: `aad-auth`
- port the i18n logic from adsys and adjust it for this project
- create FR locale and mark some strings as translatable to confirm automations work
- add GH Workflow to auto-update translation files automatically when something changes
- add `aad-common` binary package that ships translations shared between the other binary packages

Closes #29

### DIFF
--- a/debian/aad-common.install
+++ b/debian/aad-common.install
@@ -1,0 +1,1 @@
+usr/share/locale
--- a/debian/control
+++ b/debian/control
@@ -21,6 +21,8 @@ Package: libpam-aad
 Architecture: any
 Built-Using: ${misc:Built-Using},
 Depends: ${shlibs:Depends},
+Depends: aad-common,
+         ${shlibs:Depends},
          ${misc:Depends},
 Description: ${source:Synopsis} module for PAM
  ${source:Extended-Description}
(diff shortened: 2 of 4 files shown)


didrocks (commented):
Do you mind testing the PAM part in a real usage? Having some translated strings in some locals by the PAM package itself (in a module) and checking that we load our translated strings too (in the logs for instance)?

I’m puzzled because you are loading the textdomain globally, so you either erase the PAM global settings for local, or i18n.G() is using the PAM one and we don’t use our text domain. I’m happy to be prooved wrong though :p

didrocks (approved):
Thanks for testing! Indeed, I found that we store the domain in the locale object and ths is what is reused throughout the chain when calling i18n.G(). All good then, thanks for the work :)

One missing thing: you need to add a dependency between the lib* packages on aad-common. Once done, feel free to merge :)

PR #51, opened 2022-08-29
Making the linter duplicate rule more explicit.

denisonbarbosa:
Triggering the linter warns to take a look at how to handle them.


didrocks (approved):
There were not so many ones to ignore compared to the amount of the project size. Thanks, you can merge it!

PR #52, opened 2022-08-29
Replace SetShadowMode by WithShadowMode in cache tests

denisonbarbosa:
Now that the db setup inside the tests is refactored, we no longer need to rely on the setShadowMode hack in order to test some functionalities.


didrocks (approved):
Nothing to add and as we discussed, this really simplified some tests. Thanks for tackling this! Approving

PR #53, opened 2022-08-30
Removing nss testopts

denisonbarbosa:
This removes the global variable testopts from the internal/nss package and refactor the packages to use cache.Options as optional arguments for some functions that relied upon testopts. Now the tests can be executed in parallel (and they are).
Also, the nss package now uses a global variable of his own rather than the passed testopts.

### DIFF
--- a/internal/nss/group/group.go
+++ b/internal/nss/group/group.go
@@ -18,16 +18,8 @@ type Group struct {
 	members []string /* Members of the group */
 }
 
-var testopts = []cache.Option{}
-
-// setCacheOption set opts everytime we open a cache.
-// This is not compatible with parallel testing as it needs to change a global state.
-func setCacheOption(opts ...cache.Option) {
-	testopts = opts
-}
-
 // NewByName returns a passwd entry from a name.
-func NewByName(ctx context.Context, name string) (g Group, err error) {
+func NewByName(ctx context.Context, name string, cacheOpts ...cache.Option) (g Group, err error) {
 	defer func() {
 		if err != nil {
 			err = fmt.Errorf("failed to get group entry from name %q: %w", name, err)
@@ -41,7 +33,7 @@ func NewByName(ctx context.Context, name string) (g Group, err error) {
 		return Group{}, nss.ErrNotFoundENoEnt
 	}
 
-	c, err := cache.New(ctx, testopts...)
+	c, err := cache.New(ctx, cacheOpts...)
 	if err != nil {
 		return Group{}, nss.ConvertErr(err)
 	}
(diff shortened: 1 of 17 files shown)


internal/nss/group/group.go
```diff
@@ -93,12 +85,12 @@ var groupIterationCache *cache.Cache
 
 // StartEntryIteration open a new cache for iteration.
 // This needs to be called prior to calling NextEntry and be closed with EndEntryIteration.
-func StartEntryIteration(ctx context.Context) error {
+func StartEntryIteration(ctx context.Context, cacheOpts ...cache.Option) error {
```
denisonbarbosa, line 88: Since I can't comment on a line that's not in the diff, I'll add the comment right here...
The race is all about the (group|passwd|shado)IterationCache variable that we have in internal/nss/ subpackages. We have some mechanisms to avoid racing in a single execution (like this StartEntryIteration), but this does not prevent racing in parallel executions (as we've seen in the tests report). We could use a mutex to avoid this, but this is kind of problem will not happen in a real use of the lib, so we might have to do that in the tests that rely on this functionality rather than on the lib itself.

didrocks:
The code itself is good. The only remaining part is this race in the tests. As said on IRC, I left you on figuring out why with the upstream documentation :) (and then, do the changes that needs to be done to make sense in that context)

denisonbarbosa:
A little side note: The docs that I've used are [,](https://firefox-source-docs.mozilla.org/security/nss/legacy/nss_api_guidelines/index.html#mozilla-projects-nss-nss-api-guidelines) which contains some guidelines for using the NSS api.
I'm not sure if this is the best approach, since the tests were starting iterations on different databases, and if we add a mutex to solve this problem, there is no real reason to run the tests in parallel 🥲 . But also does not make sense if we change the variable to accommodate multiple tests running like this.
Anyway, I pushed it as is just so we can discuss it with some visuals.

didrocks:
Gabriel is right about NSS vs the NSS you search. I wonder how you were able to map to the current technology with the wrong man page :)

TBH, the 2 last commits should be removed. This is not what we need. Some remarks:
- "Refactoring some testcases that would result in a deadlock." makes think you did some refactoring/enhancement, while you just removed the 3 tests testing current transaction. https://github.com/ubuntu/aad-auth/commit/28a6c2bfd6508eb627f03b62cb0ddbb35159ac45. Please do not do this! Those tests are real use-case and if they can trigger a deadlock (which they don’t if the API is use correctly), then, it’s up to the code to be fixed, not the tests to be removed.
- https://github.com/ubuntu/aad-auth/commit/5899571416e63ce0f35bbfc3ec0e8003655488f4 is a good attempts to add locks and such. However, this is not what the upstream API contracts says, using the current upstream man pages for the API which are under tests:
  - https://linux.die.net/man/3/setpwent
  - https://linux.die.net/man/3/getpwent_r
And in particular, the note on the last man page states `The function getpwent_r() is not really reentrant since it shares the reading position in the stream with all other threads.`

So basically, contrary to the other NSS apis `getpwname_r` and `getpwuid_r` which are reentrant, this one can not be used in parallel due to the iterator/position we are reading.

The solution is then, for all tests dealing with the iteration (listing everything, changing the cursor position), to not run them in parallel.
Same happens with the groups/shadow.

So, reset to ignore the last couple of commits, and then, you can do the suggested modifications to the tests :)

Do not hesitate if you need any help :)

PR #54, opened 2022-08-31
Fix small packaging oopsie

GabrielNagy:
When staging hunks for the i18n commits I may have mistakenly only staged the added lines without the removal lines, so we ended up with 2 sets of Depends which is bad by all accounts, as it renders the package unbuildable.


didrocks (approved):
indeed, approving!

PR #55, opened 2022-08-31
Nss module in c

jibel:
C wrapper to call go getent
  
Following the adventure of loading a multithreaded shared library written in Go from NSS and the random crashes, we switched to a thin C wrapper that calls an executable that queries and fetch the required data. This is a temporary solution due to major limitations.


didrocks (approved):
oh waow, someone really enjoyed C!

PR #56, opened 2022-08-31
Command line to get passwd, group and shadow entries from local cache.

denisonbarbosa:
### Interface

``` aad_auth <command> [args] ```

#### Command

- ``` getent <database> [key] ```
    - database: Name of the database. We currently support 3 databases: passwd, group, shadow. It will query the respective services.
    - key is optional:
           - key can be a name or id;
           - If key is found the record corresponding to the key is returned;
           - If key is not found no record is returned and the exit status is 0;
           - If key is not specified then the entire database is listed;

#### Exit status 
If everything goes right (no input parsing issues), the command always returns 0. We don't reflect the nss return status in the returned code.

#### Debug
Logs verbosity are controlled in getent by NSS_AAD_DEBUG.

#### Output
Output is printed on stdout, and has multiple lines. The first line is the nss_status_code and errno values. Following line is the elements printed by the request, separated by ':'

{nssStatus}:{errno}
<return value 1>
<return value 2>

### CODE REVIEW

File: nss/aad_auth/main.go [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+		entries, err := GetEnt(ctx, db, key)
+
+		fmt.Print(fmtOutput(ctx, entries, err))
+	}
+}
+
+func fmtOutput(ctx context.Context, entries []string, err error) string {
+	var out string
+
+	status, errno := 1, 0
```
didrocks, line 35: Don’t hardcode 1 please, what is 1? how would I know in like 2 years?

It’s better to import the const from NSS to map it to `NSS_STATUS_SUCCESS`.

Actually errToCStatus handles the nil return values and will map it to NSS_STATUS_SUCCESS, 0, so always call it on err, even if nil?

nss/aad_auth/main.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		status, errno = errToCStatus(ctx, err)
+	}
+	out = fmt.Sprintf("%d:%d\n", status, errno)
+
+	for _, entry := range entries {
+		out += (entry + "\n")
```
didrocks, line 42: so, here, if entry is a slice of stringer, you can do:
`out = fmt.Sprintf("%s%s\n"`, out, entry)`
and you don’t even have to sweat about calling `.String()` :)

didrocks (changes requested):
I think you missed some of the things we discussed yesterday, like tests needed for the next export API in shadow/group/passwd, some comments needed on the error handling in the function, having --help printing the usage, filtering invalid inputs, copy all test cases from our existing integration tests and add the "empty db" test case…). This was for all those reasons we discussed for you to take more time before submitting for review :)

I think the general approach is correct, but here are some suggestions for a more consistent API too and testing story. Let me know what you think!

nss/aad_auth/aadauth_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"flag"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/ubuntu/aad-auth/internal/cache"
+	"github.com/ubuntu/aad-auth/internal/testutils"
+)
+
+func TestGetAllEnt(t *testing.T) {
```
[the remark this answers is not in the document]
didrocks: Another thing that we discussed yesterday was to take all the user cases/copy the integration tests cases basically and map them all, as they cover quite some large part of the code and I don’t see it?

Finally, we also mentioned to add a new use case with empty database, it’s not here if I am correct?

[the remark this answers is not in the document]
denisonbarbosa: They could be on the same test func indeed. I made them separate because the tests that only want a single entry can be executed in parallel. I thought it would be better to have this distinction... However, having to create only a single database for them would probably be faster than what we've got right now.

nss/aad_auth/error_c.go
```diff
@@ -0,0 +1,49 @@
...
+		errno = C.ENOENT
+	case err != nil: // Unexpected returned error
+		nssStatus = C.NSS_STATUS_SUCCESS
+		errno = C.EINVAL
+	}
+
+	if err != nil {
+		logger.Debug(ctx, "%v", err)
+	}
+
```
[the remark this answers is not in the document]
denisonbarbosa: We don't have the errnop* anymore. You commented this code yesterday and we did got over it real quick, so I thought there was no need for it anymore.

nss/aad_auth/aadauth_test.go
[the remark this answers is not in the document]
didrocks: good spot about parallel/non parallel.

If we want to be overkill, we can actually add a "parallel" parameter to the struct and only run t.Parallel() when needed, BUT I think this is really overkill. Let’s only gain on maintenance side.

nss/aad_auth/error_c.go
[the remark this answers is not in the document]
didrocks: We are still converting to errno and nssStatus still, no? I’m not sure to understand you here.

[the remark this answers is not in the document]
denisonbarbosa: We are, but now we return them directly in the function, rather than storing the C value in a pointer. I can just add a logger call to Debug both values, but they are covered indirectly in the fmtOutput tests

nss/aad_auth/getent.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/ubuntu/aad-auth/internal/cache"
+	"github.com/ubuntu/aad-auth/internal/logger"
+	"github.com/ubuntu/aad-auth/internal/nss"
+	"github.com/ubuntu/aad-auth/internal/nss/group"
+	"github.com/ubuntu/aad-auth/internal/nss/passwd"
+	"github.com/ubuntu/aad-auth/internal/nss/shadow"
+)
+
+// GetEnt processes the args and queries the database for the requested entries.
+func GetEnt(ctx context.Context, dbName, key string, cacheOpts ...cache.Option) (entries []string, err error) {
```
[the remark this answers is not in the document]
denisonbarbosa: I think we should rename the function again then. GetEnt should return the queried entities rather than a formatted string with the output, shouldn't it? Doesn't make much sense calling GetEnt and receiving a string with the error code and etc.

nss/aad_auth/error_c.go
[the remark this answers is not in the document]
didrocks: The idea is to not cover them in tests, but just to print something to the end user in syslogs for debugging purposes! So yes, printing both if err != nil will be of interest

nss/aad_auth/aadauth_test.go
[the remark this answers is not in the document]
didrocks: Not running the integration tests directly, but we mentioned yesterday running the *test cases*. Basically, we already have most of the use cases we want to test against and having both set of values tested on the CLI and integration tests helps. Like if both fails -> probably the issue is on the Go side. If only integration tests fails -> the issue is in the horrible C code :)

nss/aad_auth/getent.go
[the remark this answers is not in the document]
didrocks: Yes, let’s open a dictionary and pick a random name? :)

Also, you can argue the other way: GetEnt is what is the argument passed to the command line, so it’s the name, but maybe with the `Getent` case?

nss/aad_auth/internal_test.go
```diff
@@ -0,0 +1,41 @@
...
+		"properly format with unknown Err":                {err: fmt.Errorf("SomeError")},
+		"properly format with nil error":                  {err: nil},
+	}
+
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			entries := []string{"myuser@domain.com:x:1929326240:1929326240::/home/myuser@domain.com:/bin/bash"}
```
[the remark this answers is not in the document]
denisonbarbosa: Oh, now I'm commited... You'll see it.

nss/aad_auth/main.go
```diff
@@ -0,0 +1,61 @@
...
+			log.Fatalf("Request db %s is not supported", db)
+		}
+
+		out := Getent(ctx, db, key)
+		fmt.Print(out)
+	case "":
+		flag.Usage()
+		os.Exit(1)
+	default:
+		log.Fatalf("Invalid argument %s", flag.Arg(0))
```
didrocks, line 35: Same here, print the usage after the error message as the user doesn’t know how to use it :)

I suggest that you feed an err in both `case ""` and `default`, and then, out, (this will work for Getent too), you check if the err is not nil and print the error, then usage, then exit.

nss/aad_auth/mock_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+type entryMock struct {
+	str string
+}
+
+func (e entryMock) String() string {
+	return e.str
+}
+
+func getMockEntries(n int) []fmt.Stringer {
```
didrocks, line 15: This is nice :) I would just rename it to newMockEntries as you instantiate them :)

nss/aad_auth/getent.go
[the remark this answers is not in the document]
didrocks: let’s use that name :)

nss/aad_auth/getent.go
```diff
@@ -0,0 +1,154 @@
...
+		var entry fmt.Stringer
+		entry, err = nextEntryForDB(ctx, dbName)
+		if err != nil {
+			// The iteration ends with ErrNoEnt, even when it's successful.
+			break
+		}
+		entries = append(entries, entry)
+	}
+
+	if !errors.Is(err, nss.ErrNotFoundENoEnt) {
```
didrocks, line 106: Indeed, it was a good idea to remove the defer() now that the function is simpler. It’s more readable, thanks!

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_with_empty_cache
```diff
@@ -0,0 +1,2 @@
+|
```
didrocks, line 1: with empty cache cases this should return the numeric equivalent of:

status: NSS_STATUS_NOTFOUND
errno: SUCCESS

(https://www.gnu.org/software/libc/manual/html_node/NSS-Modules-Interface.html).
(Note: the bug could be in the `internal/nss/` layer)

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_without_any_cache
```diff
@@ -0,0 +1,2 @@
+|
```
didrocks, line 1: Good that we have tests :)

without any cache this should return the numeric equivalent of:

status: NSS_STATUS_UNAVAIL
errno: ENOENT

(https://www.gnu.org/software/libc/manual/html_node/NSS-Modules-Interface.html).
(Note: I’more more sure in this case that we return the right thing in `internal/nss/` layer, but worth checking. You can see this is the return value when we try to get something from shadow without having access to it).

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_passwd_without_permission_on_cache
```diff
@@ -0,0 +1,2 @@
+|
+    -1:2
```
didrocks, line 2: Here, we return the right status for instance.

nss/aad_auth/aadauth_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		// Try to list without permission on cache
+		"try to list passwd without permission on cache": {db: "passwd", rootUID: 4242},
+		"try to list group without permission on cache":  {db: "group", rootUID: 4242},
+		"try to list shadow without permission on cache": {db: "shadow", rootUID: 4242},
+	}
+
+	// Setting the DB that is not changed which will be used in most tests.
+	defaultCacheDir := t.TempDir()
+	testutils.PrepareDBsForTests(t, defaultCacheDir, "users_in_db")
```
didrocks, line 75: Even if we can’t run those tests in parallel. I suggest we use separate temporary directories for every tests, and so, a separate cache directory. (basically, have this teardown in the tests).

We don’t change this db right now, but imagine one day we start changing it with some increments for migration or any additional field we insert dynamically -> another thing to change that we can prevent by isolating what needs to be isolated.
(basically, every subtest is its own island).

That will help too with the mutliple redefinition of `cacheDir = t.TempDir()` below.

didrocks (commented):
This is still in draft, but you told me on MM that it’s ready now, so rereviewing.

Some small comments, but this is going in the right direction! I like it :)

The more annoying part is the wrong statuses returned on the "empty cache" and "no cache" tests. Do you mind having a look there? (I think you can start printing by what we return from the API).

nss/aad_auth/aadauth_test.go
```diff
@@ -0,0 +1,115 @@
...
+	for name, tc := range tests {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			uid, gid := testutils.GetCurrentUIDGID(t)
+			if tc.rootUID != 0 {
+				uid = tc.rootUID
+			}
+
+			cacheDir := defaultCacheDir
+			switch tc.cacheDB {
```
denisonbarbosa, line 86: I think it's probably better to treat the possible values explicitly, just so we don't allow tests to run with typos and anything like that (even though the 'nocache' case doesn't need to be there). I could even add a default case to stop the test if the switch falls in there, wdyt?

nss/aad_auth/aadauth_test.go
```diff
@@ -0,0 +1,115 @@
...
+			case "db_with_old_users":
+				cacheDir = t.TempDir()
+				testutils.PrepareDBsForTests(t, cacheDir, tc.cacheDB)
+			case "empty":
+				cacheDir = t.TempDir()
+				testutils.NewCacheForTests(t, cacheDir)
+			case "nocache":
+				cacheDir = t.TempDir()
+			}
```
denisonbarbosa, line 95: I think it's probably better to treat the possible values explicitly, just so we don't allow tests to run with typos and anything like that (even though the 'nocache' case doesn't need to be there). I could even add a default case to stop the test if the switch falls in there, wdyt?

didrocks: We can conciliate both:
- You could add a default: to fail
- treat the no value set == default db to only have one t.TempDir() call (as we do for a lot of tests)

nss/aad_auth/main.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+	return false
+}
+
+func aadAuthUsage() {
+	fmt.Fprintln(os.Stderr, `
+This executable should not be used directly, but should you wish too:
+
+Usage: aad_auth getenv {dbName} {key}
```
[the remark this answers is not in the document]
denisonbarbosa: Oops

nss/aad_auth/main.go
```diff
@@ -0,0 +1,61 @@
...
+		
+dbName: Name of the database to be queried. Supported values: passwd, group, shadow
```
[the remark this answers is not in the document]
denisonbarbosa: I'm not sure I understand what you mean here.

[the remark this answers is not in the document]
didrocks: Basically, do not hardcode the dbs and key in the usage string. For ease of maintainance, have all that in a map (see comment below) and reuse that same map in the code to filter supported dbs.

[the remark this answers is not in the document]
denisonbarbosa: Ack. I'm not about about treating the keys though... Since they can be strings (for querying by name), will be a pain to treat them. It's probably best to let it go on and return the output with ErrNoEnt.

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_with_empty_cache
denisonbarbosa: The problem is in internal/nss indeed. The error that is return is nss.ErrNotFoundENoEnt, but according to the documentation, it should be nss.ErrNotFoundSuccess.

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_without_any_cache
denisonbarbosa: Again, the problem is likely on the internal layer. When we call the function to [StartEntryIteration](https://github.com/ubuntu/aad-auth/blob/f98df3993b8619c5cad4bdfb4a0b5d38e568d3c8/internal/nss/group/group.go#L88), it calls cache.New(). The problem is: cache.new() also calls [initDB](https://github.com/ubuntu/aad-auth/blob/f98df3993b8619c5cad4bdfb4a0b5d38e568d3c8/internal/cache/cache.go#L204) and initDB creates the databases needed if they don't exist.
Summarizing, there is no way for us to have a "no cache" case, only "empty cache".

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_with_empty_cache
denisonbarbosa: I've managed to refactor the getAllEntries func to behave as expected, as the refactoring of the internal/nss functions would take more effort and we should probably discuss it in a peer session.
Also, the function errToCStatus might need a little review as well, since the values (status / errno) returned for ErrNotFoundSuccess are STATUS_SUCCESS / ENOENT rather than NOTFOUND / SUCCESS. (this is also included in the discussion session as well).

nss/aad_auth/main.go
```diff
@@ -22,40 +22,32 @@ func main() {
...
-    - passwd: name, uid
-    - group: name, gid
-    - shadow: name`)
+    - dbName: Name of the database to be queried.
+    - key (optional): name or uid/gid of the entry to be queried for.`)
+}
+
+func exit(status int, message string) {
+	if message != "" {
+		log.Println(message)
```
didrocks, line 49: This will print on stdout the error, you should print on stderr.

nss/aad_auth/main.go
```diff
@@ -22,40 +22,32 @@ func main() {
...
-Usage: aad_auth getenv {dbName} {key}
+Usage: aad_auth getent {dbName} {key}
 		
-dbName: Name of the database to be queried. Supported values: passwd, group, shadow
-key (optional): name or uid/gid of the entry to be queried for.
-  Supported keys for:
-    - passwd: name, uid
-    - group: name, gid
-    - shadow: name`)
+    - dbName: Name of the database to be queried.
```
didrocks, line 43: As told, you can really print the supported values here, as you have it in `supportedDbs` :) But this is not a blocker.

nss/aad_auth/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+		db, key := flag.Arg(1), flag.Arg(2)
+
+		out, err := Getent(ctx, db, key)
+		if err != nil {
+			exit(1, fmt.Sprintf("Error when trying to list %q from %s: %v", key, db, err))
+		}
+		fmt.Print(out)
+	case "":
+		exit(1, "")
```
didrocks, line 31: Here the error is rather "required argument missing".

nss/aad_auth/internal_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestFmtOutput(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string]struct {
+		nEntries int
+		err      error
+	}{
+		"properly format with error ErrTryAgainEAgain":    {err: nss.ErrTryAgainEAgain, nEntries: 1},
```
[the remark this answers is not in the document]
didrocks: This is not resolved. I did not see the edge case with one error + entries added?

nss/aad_auth/error_c.go
```diff
@@ -43,6 +43,8 @@ func errToCStatus(ctx context.Context, err error) (nssStatus, errno int) {
 
 	if err != nil {
 		logger.Debug(ctx, "Returning to NSS error: %d with errno: %d", nssStatus, errno)
+	} else {
+		logger.Debug(ctx, "Returning NSS STATUS SUCCESS (%d) with errno: %d", nssStatus, errno)
```
didrocks, line 47: No need to add the constant and the integer value, just say "Returning NSS_STATUS_SUCCESS" without "%d" and the errno.

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_without_any_cache
didrocks: We can have a test this way: if we can’t create the db.

So, the test is: "no database file and no permission to create it", mind having a shot?

nss/aad_auth/getent.go
```diff
@@ -113,11 +113,11 @@ func getAllEntries(ctx context.Context, dbName string, cacheOpts ...cache.Option
 		return nil, err
 	}
 
-	if len(entries) != 0 {
-		err = nil
+	if len(entries) == 0 {
+		return nil, nss.ErrNotFoundSuccess
```
didrocks, line 117: ack on the change, I suggest you add a comment here so that we know why the error is hardcoded.

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_with_empty_cache
didrocks: I agree that this is the best direction for now to go on the empty tests.

well, we return the status before the errno? (which doesn’t map with the go error we return)

didrocks (commented):
Still some fixes to do.

Btw, the QA linting tests is failing from the start, you should have a look there ;)

nss/aad_auth/testdata/golden/TestGetEnt/try_to_list_group_with_empty_cache
denisonbarbosa: I think so. So if I just invert the values it will be fine? In [here](https://github.com/ubuntu/aad-auth/blob/f98df3993b8619c5cad4bdfb4a0b5d38e568d3c8/nss/integration_tests/util_c.go#L38), I could change to be like:
```Go
nssStatus = C.ENOENT
errno = C.NSS_STATUS_SUCCESS
```

didrocks: No no no. What I meant is that in the Go error, we are quite logic by assessing errno, then status.

However, in C, you check the status code and then the errno. So inversing the returned value as the code does it logical from a C standpoint.

didrocks (approved):
All tests passes, no more comments. Excellent work!

Feel free to merge now :)

PR #57, opened 2022-09-01
Fixed exit status handling

jibel:
External command and nss status were mixed up. We now only consider nss
exit status. Command line exit status is used to properly set errno and
nss exit status on error.

Co-authored-by: Didier Roche <didrocks@ubuntu.com>


didrocks (approved):
Hopefully, we are done on dealing the exit status!

PR #59, opened 2022-09-05
Adapting nss integration tests

denisonbarbosa:
Adapting the tests to fit in the new NSS C library.


nss/aad-auth/helper_test.go
```diff
@@ -0,0 +1,143 @@
...
+	cflags := "-g -Wall -Wextra"
+	out, err := exec.Command("pkg-config", "--cflags", "glib-2.0").CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("could not get the required cflags (%s): %w", out, err)
+	}
+	s := string(out)
+	cflags += " " + s[:len(s)-1] // Ignoring the last \n.
+
+	// Gets the ldflags
+	out, err = exec.Command("pkg-config", "--libs", "glib-2.0").CombinedOutput()
```
didrocks, line 120: You can combine that in one call (cflags and ldflags). Add also `nss`

nss/aad-auth/helper_test.go
```diff
@@ -0,0 +1,143 @@
...
+func TestMain(m *testing.M) {
+	// Build the nss library and executable in a temporary directory and allow linking to it.
+	tmpDir, cleanup, err := createTempDir()
+	if err != nil {
+		os.Exit(1)
+	}
+	defer cleanup()
+
+	libPath = filepath.Join(tmpDir, "libnss_aad.so.2")
+	execPath = filepath.Join(tmpDir, "aad_auth")
```
didrocks, line 72: The name of the binary is `aad-auth`. Let’s not add another naming convention :)

nss/aad-auth/helper_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	testutils.InstallUpdateFlag()
+	flag.Parse()
+
+	m.Run()
+}
+
+func buildNssCLib() error {
+	tmp, err := os.ReadDir("../")
```
didrocks, line 96: maybe we could use `filepath.Glob()`? (Also, look at my comment about moving the dangerous C override file :))

didrocks (changes requested):
Some small comments, but overall, this is great!

Please double check again that all previous tests cases are still covered with the refactoring, I saw at least that those one are not: try to list by name or uid in group and passwd without access to shadow.

Also, we discussed on MM about listing the group. I doubt this is a buffer issue, we can check together.

.github/workflows/qa.yaml
```diff
@@ -38,8 +38,12 @@ jobs:
           # Build PAM library
           go build -ldflags='-s -w' -buildmode=c-shared -o pam_aad.so ./pam
 
-          # Build NSS library
-          go build -ldflags='-s -w' -buildmode=c-shared -o libnss_aad.so.2 ./nss/integration_tests
+          # Build Nss CLI executable
+          go build -o aad-auth ./nss/aad-auth
+
+          # Build NSS C library
+          cd ./nss && make libnss_aad.so.2
```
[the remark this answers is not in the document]
denisonbarbosa: Thanks for the tip! I suck when it comes to Makefile and make, that's why I use cmake instead hahaha

didrocks (approved):
All comments are resolved! Good work, it’s reassuring to have integration tests back!

PR #60, opened 2022-09-06
Aad auth various fixes

jibel:
Various fixes:

- Do not print a \n when listing entries to prevent crash
- Initialize shadow flags
- Renamed aad_auth to aad-auth
- Moved string function after New() functions


didrocks:
Ok, tests are now updated and we can log in. @denisonbarbosa FYI.

didrocks (approved):

PR #61, opened 2022-09-08
Package NSS C library

GabrielNagy:
Details in the individual commit messages

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -1,12 +1,16 @@
 #!/usr/bin/make -f
 #export DH_VERBOSE = 1
 
-export GOCACHE=$(shell mktemp -d /tmp/gocache-XXXX)
+export GOCACHE := $(shell mktemp -d /tmp/gocache-XXXX)
 export GOFLAGS := -buildmode=pie
 
-export DEB_BUILD_MAINT_OPTIONS = optimize=-lto
-# strict symbols checking
-export DPKG_GENSYMBOLS_CHECK_LEVEL=4
+export DEB_BUILD_MAINT_OPTIONS := optimize=-lto
+
+# Strict symbols checking
+export DPKG_GENSYMBOLS_CHECK_LEVEL := 4
+
+# Copy in build directory all content to embed
+export DH_GOLANG_INSTALL_ALL := 1
 
 %:
 	dh $@ --buildsystem=golang --with=golang,apport
@@ -21,10 +25,13 @@ override_dh_auto_build:
 	go build -ldflags='-extldflags -Wl,-soname,pam_aad.so' -buildmode=c-shared -o pam_aad.so ./pam
 
 	# Build NSS library
-	go build -ldflags='-extldflags -Wl,-soname,libnss_aad.so.2' -buildmode=c-shared -o libnss_aad.so.2 ./nss
+	make -C nss/
 
 	# Build the CLI
 	# DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/cmd/aad-auth dh_auto_build
+
+	# Build the NSS helper binary
+	DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/nss/aad-auth dh_auto_build
 
 override_dh_auto_install:
 	dh_auto_install -- --no-source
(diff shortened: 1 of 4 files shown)


nss/Makefile
```diff
@@ -1,11 +1,12 @@
...
+CFLAGS := -g -Wall -Wextra $(shell pkg-config --cflags glib-2.0)
+LDFLAGS := $(shell pkg-config --libs glib-2.0)
+SCRIPTPATH ?= /usr/libexec/aad-auth
 
-all: clean libnss_aad.so.2 copy
+all: clean libnss_aad.so.2
 
 libnss_aad.so.2:
-	$(CC) -DSCRIPTPATH=\"/home/u/getent.sh\" passwd.c group.c shadow.c common.c  $(CFLAGS) $(LDFLAGS) -fPIC -shared -Wl,-soname,libnss_aad.so.2 -o libnss_aad.so.2
+	$(CC) -DSCRIPTPATH=\"$(SCRIPTPATH)\" passwd.c group.c shadow.c common.c $(CFLAGS) $(LDFLAGS) -fPIC -shared -Wl,-soname,libnss_aad.so.2 -o libnss_aad.so.2
```
didrocks, line 9: could we take all *.c in the `nss/` directory? I asked for a change on dboverride to Denison to not be built, so this will have no effect.

debian/rules
```diff
@@ -21,14 +25,15 @@ override_dh_auto_build:
...
 
 	# Build NSS library
-	go build -ldflags='-extldflags -Wl,-soname,libnss_aad.so.2' -buildmode=c-shared -o libnss_aad.so.2 ./nss
+	make -C nss/
 
 	# Build the CLI
 	# DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/cmd/aad-auth dh_auto_build
 
-override_dh_auto_install:
-	dh_auto_install -- --no-source
```
didrocks, line 30: hum, we still don’t install the sources, right?

didrocks (changes requested):
Some small additions/changes to do, but otherwise, good work as usual :)

conf/aad.conf.template
```diff
@@ -1,16 +1,23 @@
-# required values
-tenant_id = <TENANT_ID>
-app_id = <APPLICATION OR CLIENT ID>
+### required values
```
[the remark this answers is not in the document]
didrocks: ah, that were the changes in your CLI branch. Got it :) So yeah, we took it

debian/rules
didrocks: Ack then!

debian/rules
```diff
@@ -21,26 +25,31 @@ override_dh_auto_build:
...
+	# Install configuration template to /etc
 	mkdir -p debian/tmp/etc
-	sed -e 's/^\(\)/# \1/' conf/aad.conf.template > debian/tmp/etc/aad.conf
+	cp conf/aad.conf.template debian/tmp/etc/aad.conf
 
 	# Install PAM and NSS libraries
 	mkdir -p debian/tmp/usr/lib/$(DEB_HOST_GNU_TYPE)/security
 	cp pam_aad.so debian/tmp/usr/lib/$(DEB_HOST_GNU_TYPE)/security
-	cp libnss_aad.so.2 debian/tmp/usr/lib/$(DEB_HOST_GNU_TYPE)
+	cp nss/libnss_aad.so.2 debian/tmp/usr/lib/$(DEB_HOST_GNU_TYPE)
```
didrocks, line 48: We should use DESTDIR so that make install install it in the correct directory.

didrocks: yes, taking DESTDIR to be as standard as possible.

debian/rules
```diff
@@ -21,26 +25,31 @@ override_dh_auto_build:
...
-	go build -ldflags='-extldflags -Wl,-soname,libnss_aad.so.2' -buildmode=c-shared -o libnss_aad.so.2 ./nss
+	make -C nss/
 
 	# Build the CLI
 	# DH_GOLANG_BUILDPKG=github.com/ubuntu/aad-auth/cmd/aad-auth dh_auto_build
 
-override_dh_auto_install:
-	dh_auto_install -- --no-source
+	# Build the NSS helper binary
+	go build ./nss/aad-auth
```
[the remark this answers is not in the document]
didrocks: hum, they do set lto for instance, I wonder if we can reuse some parts of it. Also, we don’t pass CFLAGS/LDFLAGS passed by the builder for instance.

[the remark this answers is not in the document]
didrocks: yeah, that would be the best!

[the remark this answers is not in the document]
didrocks: ack!

debian/rules
didrocks: No, this looks good to me. Would have been great to have some more dh_ calls here, but I guess anyway that for this frankeinstein project, we will have a different approach with Rust :)

didrocks (approved):
Looks good! let’s get it in so that tomorrow, we can cut a release :)

PR #64, opened 2022-09-12
Improve consistency of offline credentials expiration

denisonbarbosa:
Now the offline_credentials_expiration value can be less than zero. This value will disable offline authentication and default the cache purge policy to 180 days.

This means that the 0 value now works as intended and correctly prevents the cache clean up process, allowing offline login for an undetermined amount of time.

DEENG-389


internal/cache/cache.go
```diff
@@ -208,9 +210,17 @@ func New(ctx context.Context, opts ...Option) (c *Cache, err error) {
...
+			d = defaultCredentialsExpiration
+		}
+		days := uint64(d) * expirationPurgeModifier
+
+		logger.Debug(ctx, "Cleaning up db. Removing entries that last authenticated online more than %d days ago", days)
+
+		maxCacheEntryDuration := time.Duration(days * 24 * uint64(time.Hour))
+		if err := cleanUpDB(ctx, db, maxCacheEntryDuration); err != nil {
 			return nil, err
 		}
```
didrocks, line 225: I would put an else here for debug telling that we don’t clean up the database because offlineCredentialsExpiration is that to 0, as the one you removed in `CleanUpDB`.

internal/cache/cache.go
```diff
@@ -208,9 +210,17 @@ func New(ctx context.Context, opts ...Option) (c *Cache, err error) {
...
-		offlineCredentialsExpirationDuration := time.Duration(2 * uint64(o.offlineCredentialsExpiration) * 24 * uint64(time.Hour))
-		if err := cleanUpDB(ctx, db, offlineCredentialsExpirationDuration); err != nil {
+	if shadowMode == shadowRWMode && o.offlineCredentialsExpiration != 0 {
+		d := o.offlineCredentialsExpiration
+		if o.offlineCredentialsExpiration < 0 {
+			d = defaultCredentialsExpiration
+		}
+		days := uint64(d) * expirationPurgeModifier
+
+		logger.Debug(ctx, "Cleaning up db. Removing entries that last authenticated online more than %d days ago", days)
```
didrocks, line 220: I would really add this log to `cleanUpDB` directly (I know you added it here because of the number of days, but you can at worst convert it back, it really belongs in cleanUpDB IMHO).

internal/cache/cache.go
```diff
@@ -301,15 +311,22 @@ func (c *Cache) CanAuthenticate(ctx context.Context, username, password string)
...
 		return err
 	}
 
 	// ensure that we checked credential online recently.
-	logger.Debug(ctx, "Last online login was: %s. Current time: %s. Revalidation needed every %d days", user.LastOnlineAuth, time.Now(), c.offlineCredentialsExpiration)
-	if time.Now().After(user.LastOnlineAuth.Add(time.Duration(uint64(c.offlineCredentialsExpiration) * 24 * uint64(time.Hour)))) {
-		return ErrOfflineCredentialsExpired
+	logger.Debug(ctx, "Last online login was: %s. Current time: %s.", user.LastOnlineAuth, time.Now())
+	if c.offlineCredentialsExpiration > 0 {
+		logger.Debug(ctx, "Revalidation needed every %d days", c.offlineCredentialsExpiration)
```
didrocks, line 326: "Online revalidation"? wdyt?

internal/cache/cache_test.go
```diff
@@ -195,14 +195,16 @@ func TestCleanupDB(t *testing.T) {
...
 
 	tests := map[string]struct {
 		offlineCredentialsExpirationTime *int
 
 		wantKeepOldUsers bool
 	}{
-		"clean up old users":     {},
-		"do not clean up anyone": {offlineCredentialsExpirationTime: &zeroDuration, wantKeepOldUsers: true},
+		"clean up old users": {},
+		"clean up old users when offline auth is disabled": {offlineCredentialsExpirationTime: &offlineAuthDisabled},
```
didrocks, line 206: I would clarify the use case name like "with our default cleanup policy"

internal/cache/cache_test.go
```diff
@@ -345,21 +347,23 @@ func TestCanAuthenticate(t *testing.T) {
...
 		"can authenticate even with shadow file RO":     {userPasswords: map[string]string{"myuser@domain.com": "my password"}, shadowMode: &cache.ShadowROMode},
+		"can authenticate even with very old user":      {userPasswords: map[string]string{"veryolduser@domain.com": "my password"}, useoldaccounts: true},
 
 		// error cases
 		"error on wrong password":                         {userPasswords: map[string]string{"myuser@domain.com": "wrong password"}, wantErr: true},
 		"error on wrong user":                             {userPasswords: map[string]string{"does not exist user": "my password"}, wantErr: true},
 		"error on checking when can’t access shadow file": {userPasswords: map[string]string{"myuser@domain.com": "my password"}, shadowMode: &cache.ShadowNotAvailableMode, wantErr: true},
-		"do not let too old unpurged accounts to log in ": {userPasswords: map[string]string{"veryolduser@domain.com": "my password"}, useoldaccounts: true, wantErr: true},
+		"error on offline authentication disabled":        {userPasswords: map[string]string{"myuser@domain.com": "my password"}, wantErr: true, disabledOfflineAuth: true},
```
didrocks, line 366: wantErr should be after the function input (we have the inputs, then the wants)

internal/cache/cache_test.go
```diff
@@ -345,21 +347,23 @@ func TestCanAuthenticate(t *testing.T) {
...
 		"handle separately multiple users and password": {userPasswords: map[string]string{"myuser@domain.com": "my password", "otheruser@domain.com": "other password"}},
 		"can authenticate even with shadow file RO":     {userPasswords: map[string]string{"myuser@domain.com": "my password"}, shadowMode: &cache.ShadowROMode},
+		"can authenticate even with very old user":      {userPasswords: map[string]string{"veryolduser@domain.com": "my password"}, useoldaccounts: true},
 
 		// error cases
 		"error on wrong password":                         {userPasswords: map[string]string{"myuser@domain.com": "wrong password"}, wantErr: true},
 		"error on wrong user":                             {userPasswords: map[string]string{"does not exist user": "my password"}, wantErr: true},
 		"error on checking when can’t access shadow file": {userPasswords: map[string]string{"myuser@domain.com": "my password"}, shadowMode: &cache.ShadowNotAvailableMode, wantErr: true},
-		"do not let too old unpurged accounts to log in ": {userPasswords: map[string]string{"veryolduser@domain.com": "my password"}, useoldaccounts: true, wantErr: true},
```
didrocks, line 362: We don’t have any test left for old unpurged account not being able to log in (meaning, accounts with last login time between 2*expiration and expiration)?

internal/cache/cache_test.go
```diff
@@ -388,6 +396,10 @@ func TestCanAuthenticate(t *testing.T) {
 					if username == "veryolduser@domain.com" {
```
didrocks, line 396: Due to the test remove, is this code still in use?

didrocks (changes requested):
Good work! I really prefer the new behavior and it seems more in line with what we want in the long term.

I have some remarks and questions that you can see above on the behaviour and I think, on some newly missed tests cases :)

internal/cache/cache_test.go
didrocks: This test is still removed, no?

didrocks: I agree that "error on offline authentication disabled" is going to the same code path, but here, we are using with an user which shouldn’t be purged with the default policy. I think it’s good to test here with an user which should have been purged too (so an additional test).

internal/pam/pam_test.go
```diff
@@ -31,21 +31,24 @@ func TestAuthenticate(t *testing.T) {
...
+		"error on unexisting conf":                              {conf: "doesnotexist.conf", wantErrType: pam.ErrPamSystem},
+		"error on unexisting users":                             {username: "no such user", wantErrType: pam.ErrPamAuth},
+		"error on invalid password":                             {username: "invalid credentials", wantErrType: pam.ErrPamAuth},
+		"error on offline with user online user not in cache":   {conf: "forceoffline.conf", initialCache: "db_with_expired_users", wantErrType: pam.ErrPamAuth},
+		"error on offline with expired user":                    {conf: "forceoffline.conf", initialCache: "db_with_expired_users", username: "expireduser@domain.com", wantErrType: pam.ErrPamAuth},
+		"error on offline with purged user":                     {conf: "forceoffline-expire-right-away.conf", initialCache: "db_with_expired_users", username: "purgeduser@domain.com", wantErrType: pam.ErrPamAuth},
+		"error on offline with offline authentication disabled": {conf: "forceoffline-offline-auth-disabled.conf", initialCache: "users_in_db", username: "myuser@domain.com", wantErrType: pam.ErrPamAuth},
```
didrocks, line 49: same comment that the other comment on the removed tests which is slightly different, even if the same code path.

pam/integration_test.go
```diff
@@ -55,9 +58,9 @@ func TestPamSmAuthenticate(t *testing.T) {
 		"error on invalid password":                           {username: "invalid credentials", wantErr: true},
 		"error on config values only in mismatching domain":   {username: "success@otherdomain.com", conf: "with-domain.conf", wantErr: true},
 		"error on offline with user online user not in cache": {conf: "forceoffline.conf", offline: true, initialCache: "db_with_expired_users", wantErr: true},
-		"error on offline with purged user account":           {username: "purgeduser@domain.com", offline: true, initialCache: "db_with_expired_users", wantErr: true},
-		"error on offline connecting expired user from cache": {conf: "forceoffline.conf", offline: true, initialCache: "db_with_expired_users", username: "expireduser@domain.com", wantErr: true},
-		"error on offline with unpurged old user account":     {conf: "forceoffline-expire-right-away.conf", offline: true, initialCache: "db_with_expired_users", username: "purgeduser@domain.com", wantErr: true},
```
didrocks, line 60: and ditto :)

didrocks (commented):
3 remaining comments which are in reality just one on the test equivalence on what was removed compared to what was added.

Otherwise, the rest looks good to me, and it’s good to be on a better base thanks to the other PR. Thanks for working on this!

internal/cache/cache_test.go
denisonbarbosa: Indeed, the test is missing. But it's because we can no longer test it, I guess. If the credentials expiration is set to 0, the user is supposed to be able to authenticate despite how long he re-validated online... So we either have a case in which the user is purged (which is the case "error on offline with purged user") or a case in which credentials do not expire (which is the case "offline, connect purged user from cache").

didrocks (approved):
After online discussion on MM, we can’t encounter the removed case. Also, coverage is still covering what was expected. Sounds then good to me and I’m happy to ACK for merging.

Nice work again! :)

PR #65, opened 2022-09-13
Improving old users dump (Should be merged before PR#64) 

denisonbarbosa:
Changing the db_with_old_users dump to db_with_expired_users. Now the last_online_auth is loaded as a wildcard value that is parsed according to the following:

- PURGE_TIME = NOW - (2 * DEF + C)
- EXPIRATION_TIME = NOW - (DEF + C)
- FUTURE_TIME = NOW + C
- RECENT_TIME = NOW - C                           // This already existed, but I thought I should mention it here as well.

-- NOW is the current time (in case there were any doubts about it);
-- DEF is our cache.defaultCredentialsExpiration (defined in cache.go), and exported in cache/export_testdb.go, currently its value is 90;
-- C is just a constant, currently its value is 2;

This allows us to have a little more versatility in the tests, since now we can test cases that we weren't testing before and removes the redundant db dump that we had before, which basically consisted of:
- a time traveler;
- a grandpa that used the system 30 years ago;
- a great grandpa that used the system 50 years ago;

We still have the time traveler, but the grandpas were replaced by more recent users that fall into two categories that we can have in the cache: _last logged in more than twice the expiration time_ and _last logged in more than the expiration time, but less than twice its value_.


internal/pam/pam_test.go
```diff
@@ -40,9 +40,10 @@ func TestAuthenticate(t *testing.T) {
...
-		"error on offline with purged user account":           {username: "veryolduser@domain.com", initialCache: "db_with_old_users", wantErrType: pam.ErrPamAuth},
-		"error on offline with unpurged old user account":     {conf: "forceoffline-expire-right-away.conf", initialCache: "db_with_old_users", username: "veryolduser@domain.com", wantErrType: pam.ErrPamAuth},
+		"error on offline with user online user not in cache": {conf: "forceoffline.conf", initialCache: "db_with_expired_users", wantErrType: pam.ErrPamAuth},
+		"error on offline with purged user account":           {username: "purgeduser@domain.com", initialCache: "db_with_expired_users", wantErrType: pam.ErrPamAuth},
+		"error on offline with unpurged old user account":     {conf: "forceoffline-expire-right-away.conf", initialCache: "db_with_expired_users", username: "purgeduser@domain.com", wantErrType: pam.ErrPamAuth},
+		"error on offline with expired user account":          {conf: "forceoffline.conf", initialCache: "db_with_expired_users", username: "expireduser@domain.com", wantErrType: pam.ErrPamAuth},
```
didrocks, line 46: nitpick (because there had to be one :p): in the previous tests you were doing expired, then unpurged, which is the right order IMHO, Please try follow the same order here.

pam/integration_test.go
```diff
@@ -41,8 +41,8 @@ func TestPamSmAuthenticate(t *testing.T) {
 		"correctly set homedir and shell values specified at domain for a new user with matching domain": {conf: "aad-with-homedir-and-shell-domain.conf"},
 
 		// offline cases
-		"offline, connect existing user from cache":                                     {conf: "forceoffline.conf", offline: true, initialCache: "db_with_old_users", username: "futureuser@domain.com"},
-		"homedir and shell values should not change for user that was already on cache": {conf: "forceoffline-with-homedir-and-shell.conf", offline: true, initialCache: "db_with_old_users", username: "futureuser@domain.com"},
+		"offline, connect existing user from cache":                                     {conf: "forceoffline.conf", offline: true, initialCache: "db_with_expired_users", username: "futureuser@domain.com"},
+		"homedir and shell values should not change for user that was already on cache": {conf: "forceoffline-with-homedir-and-shell.conf", offline: true, initialCache: "db_with_expired_users", username: "futureuser@domain.com"},
```
didrocks, line 45: ditto here.

didrocks (changes requested):
Very well done! It’s something that I wanted to do when you start adding WILDCARD support in timestamp, but hadn’t the time for. This is nice for you to have proposed and implemented it :)

Some changes requested and I think we can by this PR, replaces 2 cases where we used this db instead of the other one but where we can now use the normal one, as you are treating RECENT_TIME there.

cmd/aad-cli/cli/user_test.go
```diff
@@ -102,13 +104,21 @@ func TestUser(t *testing.T) {
...
-			}
 
+				if slices.Contains(args, "last_online_auth") {
+					i, err := strconv.ParseInt(strings.Trim(got, "\n"), 10, 64)
+					require.NoError(t, err, "Expected no error but got one")
+					require.Equal(t, user.LastOnlineAuth.Unix(), i)
+					return
+				}
+
+				if len(args) < 4 {
```
didrocks, line 115: I guess this is for `get_user*`. A little bit like the previous one, I’m not a big fan of not having golden files for those use case and having this only checked in the tests (this test now is a mix of golden and non golden file). Let’s see what @GabrielNagy think.

cmd/aad-cli/cli/user_test.go
```diff
@@ -145,12 +155,19 @@ func TestUserSetAttribute(t *testing.T) {
...
+			cacheDB := "db_with_expired_users"
 			testutils.PrepareDBsForTests(t, cacheDir, cacheDB)
 			cache := testutils.NewCacheForTests(t, cacheDir)
 			c := cli.New(cli.WithCache(cache), cli.WithCurrentUser("futureuser@domain.com"))
 
-			_, err := testutils.RunApp(t, c, args...)
+			// Gets the user time before running the cli
+			var wantTime time.Time
+			aux, err := cache.GetUserByName(context.Background(), username)
+			if err == nil {
```
didrocks, line 166: It seems the only case the user does not exist is with "no user". So I suggest that you populate the time if `username != nouser@domain.com` and assert that there is no error while trying to retrieve the time. (this is to avoid false positive if the test pass without having a time where we didn’t expect it)

internal/cache/helper_test.go
```diff
@@ -4,13 +4,14 @@ type userInfos struct {
 	name     string
 	uid      int64
 	password string
+	gecos    string
```
didrocks, line 7: Thanks for adding the gecos! (However, this could have been a separated PR to really see the golden noise diff more easily)

internal/testutils/db.go
```diff
@@ -261,8 +261,13 @@ func loadDumpIntoDB(t *testing.T, dumpPath, dbPath string) {
 			var s string
 			// Looping through the columns to ensure that the values will be ordered as supposed to.
 			for i, col := range table.Cols {
-				values[i] = row[col]
-				if col == "last_online_auth" && values[i] == "RECENT_TIME" {
+				if col == "last_online_auth" {
+					values[i] = ParseTimeWildcard(row[col]).Unix()
+				} else {
+					values[i] = row[col]
```
didrocks, line 267: I would put is before the if, then override if it’s last_online_auth and thus, removing the else clause.

didrocks (changes requested):
This is getting close to be merge. I want @GabrielNagy's opinion on my remark and see how we can do better if possible. Then, the rest looks good to me, even if I would have separated the gecos update change so that every PR stays on one topic.

I have a couple of comments that I wrote above, but this cleanup is great to see, thanks!

cmd/aad-cli/cli/user_test.go
denisonbarbosa: I agree that it would be better to print the values in the same way. And it does make our life easier as well, I'll do that.

internal/cache/helper_test.go
denisonbarbosa: Yeah... I didn't like the amount of noise diff that was generated, but there was a golden file that was empty because the users_in_db didn't have any gecos value, so I thought we might as well just change everything and deal with the noise anyway, specially since I'm switching the db that is used is some tests.

internal/testutils/db.go
denisonbarbosa: I've also realized that I forgot to remove the previous code there as well 🤡

didrocks (approved):
Ah, this is way better IMHO! Thanks for doing the changes, I have no more remarks left. I also like that we do not have the magical 0 time value either :)

PR #119, opened 2023-02-23
Updating GitHub actions

denisonbarbosa:
Rearranging the workflow actions to group them by purpose rather than by language. Before we had a QA for Rust and one for Go, now we only have only action responsible for doing the QA (for both langs) and one specific for the tests and coverage (also for both langs).

DEENG-606


didrocks:
This looks good to me! I think we should have now code coverage enabled and see, as we are using twice the same file. Can you look at the code coverage result?

denisonbarbosa:
> This looks good to me! I think we should have now code coverage enabled and see, as we are using twice the same file. Can you look at the code coverage result?

As we talked during the meeting, we are not using the same file to write the coverage. Indeed, the coverage is enabled (despite codecov weirdly not commenting it on the PR) and it can be viewed on https://app.codecov.io/gh/ubuntu/aad-auth/tree/updating-github-actions/. However, Rust coverage is way too buggy and weird to evaluate, so the end result is a bit off.
My guess is that we will have more luck using the same approach as we used in adsys, but then we will lose some covered lines I think.
Also, we should probably upload the coverage files as artifacts to the repository in order to diff them before submitting a new coverage report to codecov (maybe this will help prevent the inconsistent diff reports that often happen in adsys), wdyt @didrocks ?

didrocks:
> Also, we should probably upload the coverage files as artifacts to the repository in order to diff them before submitting a new coverage report to codecov (maybe this will help prevent the inconsistent diff reports that often happen in adsys), wdyt @didrocks ?

I’m unsure what this would give us. The coverage file will probably be different as the PR is supposively changing the files, doesn't it?

Do you mind looking at the comment part, if there is any opened bugs and so so that we understand what happens? I agree the numbers are not really great. I suggest that we also look if there is a way to skip some files, like `_test.rs` one.

denisonbarbosa:
> I’m unsure what this would give us. The coverage file will probably be different as the PR is supposively changing the files, doesn't it?
Honestly, I'm not sure anymore. I've tried running it multiple times locally and it always gave me the same results. However, now that I think of it, trying to diff the coverage artifacts is quite a fragile idea since it would report a difference even if the lines changed in order.
> I suggest that we also look if there is a way to skip some files, like `_test.rs` one.
Yeah, it was quite easy to skip the tests and the testutils files, just had to fix the regex.

> Do you mind looking at the comment part, if there is any opened bugs and so so that we understand what happens? I agree the numbers are not really great. 
So, about that... I found a couple of reports of the same issue, but the fixes seem to differ. I would like us to first check the Codecov token of the repo and also take a look at [this link](https://about.codecov.io/blog/codecov-is-updating-its-github-integration/). I think I do not have enough repo permissions to handle it, so I'll need help from one who has.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,70 @@
...
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v3
+        - uses: actions/setup-go@v3
+          with:
+            go-version-file: go.mod
+            cache: true
+        - name: Install dependencies
+          run: |
+            sudo DEBIAN_FRONTEND=noninteractive apt update
```
didrocks, line 22: this needs `set -eu`

didrocks (approved):
One minor comment, and then, good to merge!

PR #125, opened 2023-02-24
Moving log init call to a constructor in lib.rs

denisonbarbosa:
The logger was being initialized in each hook call, which could result in races when running the tests (which run in parallel by default) or when executing multiple NSS calls in sequence.

Now, after creating the constructor and removing the initialization from each hook function, this issue should be prevented since the logger will only be initialized when the library is being initialized.


didrocks (approved):
So, it seems that we can have multiple constructors, great!

PR #132, opened 2023-03-01
Add coverage to NSS integration tests

denisonbarbosa:
Reusing the tools created in adsys to track Rust coverage when running Go tests and adapting the integration tests to the new build pattern.

DEENG-613

### DIFF
--- a/internal/testutils/coverage.go
+++ b/internal/testutils/coverage.go
@@ -1,0 +1,158 @@
+package testutils
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/slices"
+)
+
+var (
+	goMainCoverProfile     string
+	goMainCoverProfileOnce sync.Once
+
+	coveragesToMerge   []string
+	coveragesToMergeMu sync.Mutex
+)
+
+// TrackTestCoverage starts tracking coverage in a dedicated file based on current test name.
+// This file will be merged to the current coverage main file.
+// It’s up to the test use the returned path to file golang-compatible cover format content.
+// To collect all coverages, then MergeCoverages() should be called after m.Run().
+// If coverage is not enabled, nothing is done.
+func TrackTestCoverage(t *testing.T) (testCoverFile string) {
+	t.Helper()
+
+	goMainCoverProfileOnce.Do(func() {
+		for _, arg := range os.Args {
+			if !strings.HasPrefix(arg, "-test.coverprofile=") {
+				continue
+			}
+			goMainCoverProfile = strings.TrimPrefix(arg, "-test.coverprofile=")
+		}
+	})
+
+	if goMainCoverProfile == "" {
+		return ""
+	}
+
+	coverAbsPath, err := filepath.Abs(goMainCoverProfile)
+	require.NoError(t, err, "Setup: can't transform go cover profile to absolute path")
+
+	testCoverFile = fmt.Sprintf("%s.%s", coverAbsPath, strings.ReplaceAll(
+		strings.ReplaceAll(t.Name(), "/", "_"),
+		"\\", "_"))
+
+	coveragesToMergeMu.Lock()
+	defer coveragesToMergeMu.Unlock()
+	if slices.Contains(coveragesToMerge, testCoverFile) {
+		t.Fatalf("Trying to adding a second time %q to the list of file to cover. This will create some overwrite and thus, should be only called once", testCoverFile)
+	}
+	coveragesToMerge = append(coveragesToMerge, testCoverFile)
+
+	return testCoverFile
+}
+
+// MergeCoverages append all coverage files marked for merging to main Go Cover Profile.
+// This has to be called after m.Run() in TestMain so that the main go cover profile is created.
+// This has no action if profiling is not enabled.
+func MergeCoverages() {
+	coveragesToMergeMu.Lock()
+	defer coveragesToMergeMu.Unlock()
+	for _, cov := range coveragesToMerge {
+		fmt.Printf("COVERAGE: %s", cov)
+		if err := appendToFile(cov, goMainCoverProfile); err != nil {
+			log.Fatalf("Teardown: can’t inject coverage into the golang one: %v", err)
+		}
+	}
+	coveragesToMerge = nil
+}
+
+// WantCoverage returns true if coverage was requested in test.
+func WantCoverage() bool {
+	for _, arg := range os.Args {
+		if !strings.HasPrefix(arg, "-test.coverprofile=") {
+			continue
+		}
+		return true
+	}
+	return false
+}
+
+// appendToFile appends src to the dst coverprofile file at the end.
+func appendToFile(src, dst string) error {
+	f, err := os.Open(filepath.Clean(src))
+	if err != nil {
+		return fmt.Errorf("can't open coverage file: %w", err)
+	}
+	defer f.Close()
+
+	d, err := os.OpenFile(dst, os.O_APPEND|os.O_WRONLY, 0600)
+	if err != nil {
+		return fmt.Errorf("can't open golang cover profile file: %w", err)
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		if strings.HasPrefix(scanner.Text(), "mode: ") {
+			continue
+		}
+		if _, err := d.Write([]byte(scanner.Text() + "\n")); err != nil {
+			return fmt.Errorf("can't write to golang cover profile file: %w", err)
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return fmt.Errorf("error while scanning golang cover profile file: %w", err)
+	}
+	return nil
+}
+
+// fqdnToPath allows to return the fqdn path for this file relative to go.mod.
+func fqdnToPath(t *testing.T, path string) string {
+	t.Helper()
+
+	srcPath, err := filepath.Abs(path)
+	require.NoError(t, err, "Setup: can't calculate absolute path")
+
+	d := srcPath
+	for d != "/" {
+		f, err := os.Open(filepath.Clean(filepath.Join(d, "go.mod")))
+		if err != nil {
+			d = filepath.Dir(d)
+			continue
+		}
+		defer func() { assert.NoError(t, f.Close(), "Setup: can’t close go.mod") }()
+
+		r := bufio.NewReader(f)
+		l, err := r.ReadString('\n')
+		require.NoError(t, err, "can't read go.mod first line")
+		if !strings.HasPrefix(l, "module ") {
+			t.Fatal(`Setup: failed to find "module" line in go.mod`)
+		}
+
+		prefix := strings.TrimSpace(strings.TrimPrefix(l, "module "))
+		relpath := strings.TrimPrefix(srcPath, d)
+		return filepath.Join(prefix, relpath)
+	}
+
+	t.Fatal("failed to find go.mod")
+	return ""
+}
+
+// writeGoCoverageLine writes given line in go coverage format to w.
+func writeGoCoverageLine(t *testing.T, w io.Writer, file string, lineNum, lineLength int, covered string) {
+	t.Helper()
+
+	_, err := w.Write([]byte(fmt.Sprintf("%s:%d.1,%d.%d 1 %s\n", file, lineNum, lineNum, lineLength, covered)))
+	require.NoErrorf(t, err, "Teardown: can't write a write to golang compatible cover file : %v", err)
+}
--- a/internal/testutils/coverage.go
+++ b/internal/testutils/coverage.go
@@ -2,9 +2,9 @@ package testutils
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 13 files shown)


nss/src/coverage.go
```diff
@@ -0,0 +1,4 @@
+// Package coveragesrc file is only here so that it’s recognized as a go package when computing coverage
+package coveragesrc
+
+import "C"
```
denisonbarbosa, line 4: There's a .c file in the src/ directory and Go was complaining about it, so I had to add this `import "C"` clausure to make it stop.

nss/src/testutils/coverage.go
```diff
@@ -0,0 +1,2 @@
+// Package coveragetestutils file is only here so that it’s recognized as a go package when computing coverage
+package coveragetestutils
```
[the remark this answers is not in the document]
denisonbarbosa: I'll try it out

nss/src/shadow/coverage.go
```diff
@@ -0,0 +1,2 @@
+// Package coverageshadow file is only here so that it’s recognized as a go package when computing coverage
+package coverageshadow
```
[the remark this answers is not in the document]
denisonbarbosa: Yeah, I tried it with underscores at the first time as well... :smiling_face_with_tear:

nss/src/coverage.go
denisonbarbosa: This is not a lint problem, I think. I get an error saying: 
> C source files not allowed when not using cgo or SWIG: db_override.c `go list`

nss/integration-tests/helper_test.go
```diff
@@ -44,44 +48,45 @@ func outNSSCommandForLib(t *testing.T, rootUID, rootGID, shadowMode int, cacheDi
...
-	}
-	targetDir = filepath.Join(targetDir, os.Getenv("DEB_HOST_RUST_TYPE"))
 
-	// Renames the compiled library to have the expected versioned name.
-	if err = os.Rename(filepath.Join(targetDir, "debug", "libnss_aad.so"), libPath); err != nil {
-		return fmt.Errorf("Setup: could not rename the Rust NSS library: %w", err)
+	out, err := cmd.CombinedOutput()
+	require.NoError(t, err, "Setup: could not build Rust NSS library: %s", out)
+
+	if cargo != "cargo" {
```
didrocks, line 81: hum, maybe tell this is to differentiate the real binary from the debian wrapper  (and so, move your comment above). How does it relate to the current PR though?

nss/integration-tests/integration_test.go
```diff
@@ -153,23 +155,22 @@ func TestIntegration(t *testing.T) {
...
 
+	if testutils.WantCoverage() {
+		defer func() {
+			// Gets the path to the integration-tests.
+			_, p, _, _ := runtime.Caller(0)
+			currDir := filepath.Dir(p)
+			// Purge .profraw files
+			files, _ := filepath.Glob(filepath.Join(currDir, "*.profraw"))
+			for _, f := range files {
+				os.Remove(f)
```
didrocks, line 169: you probably want to tell explicitely we are ignoring the error returned by Remove() like `_ = os.Remove()`

nss/integration-tests/integration_test.go
```diff
@@ -153,23 +155,22 @@ func TestIntegration(t *testing.T) {
...
+			}
+		}()
+	}
+
 	m.Run()
```
didrocks, line 174: ah, here is an issue, you should capture the return value of m.Run(), and error out with the correct exit code after collecting the coverage.

You should find some examples in other projects (or even this one), for this.

Here, if the test fails, you will always return success (and yes, it’s already there and I should have spotted it before, sorry)

nss/src/shadow/coverage.go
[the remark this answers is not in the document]
didrocks: this is annoying :p, but yeah, I don’t have immediate better ideas too.

didrocks (changes requested):
Nice work! Some suggested improvements, questions and a pre-existing bug to be fixed. Lot of diverse things to do then :)

nss/integration-tests/helper_test.go
denisonbarbosa: This was already there. I think it's just a diff weirdness, since I shifted a couple of things around to make it more clear

nss/integration-tests/integration_test.go
denisonbarbosa: Well, we should probably also fix adsys then 😂

nss/integration-tests/integration_test.go
```diff
@@ -166,11 +166,13 @@ func TestMain(m *testing.M) {
...
+				_ = os.Remove(f)
 			}
 		}()
 	}
 
-	m.Run()
+	code := m.Run()
 	testutils.MergeCoverages()
+
+	os.Exit(code)
```
didrocks, line 177: Imagine testutils.MergeCoverages() skip the test (for whatever reason), it’s better to defer os.Exit() just after m.Run() IMHO.

nss/integration-tests/helper_test.go
didrocks: I think it worthes a comment as least (because if you don’t know `dh_cargo` is using a wrapper script with a different path, you will not know why this check…)

denisonbarbosa: I agree, but the more I think about it, maybe it's better to remove this if and just leave the filepath.Join line (with some improvements on the rationale comment), since the CARGO_PATH env could still be used to set another cargo path than the one present in PATH.

nss/integration-tests/integration_test.go
denisonbarbosa: There are some problems in the way we implemented the MergeCoverages though... There are some cases where we call `log.Fatalf()` and, from what I've seen, these calls completely ignore the defers that were queued in `TestMain`, so we should probably change that as well later (or it could be on this PR as well), wdyt?

didrocks: We introduce a copy of this code here, so I think it’s part of this PR to fix it.
You are correct that if we call log.Fatalf(), we have an issue (due to os.Exit() being called). Maybe we should return an error code then? The end result would be the same, but we at least know that this can stop early.

denisonbarbosa:
> * according to the last coverage report we still seem to be missing the rust coverage https://app.codecov.io/github/ubuntu/aad-auth/commit/0767150d4e99eaa9c217fab8bf44fe175c6fddce - would be good to check if this is on us or codecov (by uploading the coverage report as an artifact in the GH Actions run, or just running `cat` on it before uploading it)

I'm so puzzled about that as well... Locally it works perfectly, but codecov ignores it for some reason. I've added the option to upload the coverage as an artifact so we can debug it.

> * no idea how to do this, but it'd be good to _not_ fail CI if the covered % decreases

It seems like a configuration from the new codecov app, I think. I don't know if we can override it through the action or if it needs to be configured in the codecov page itself, I'll have a look (appreciate any help though).

nss/integration-tests/integration_test.go
denisonbarbosa: Yeah, I think it's better. I'll update it.

nss/integration-tests/integration_test.go
```diff
@@ -153,23 +156,30 @@ func TestIntegration(t *testing.T) {
...
+	if testutils.WantCoverage() {
+		// Gets the path to the integration-tests.
+		_, p, _, _ := runtime.Caller(0)
+		currDir := filepath.Dir(p)
+		// Purge .profraw files
+		files, _ := filepath.Glob(filepath.Join(currDir, "*.profraw"))
+		for _, f := range files {
+			_ = os.Remove(f)
+		}
+	}
```
denisonbarbosa, line 184: Since the defer calls go on a stack, I had to update this scope as well to ensure that the files will be cleaned correctly

didrocks:
> I would prefer this, as there's just a few repositories using codecov (mostly us), and I think we want the same behavior everywhere. WDYT @didrocks ?

I pasted, sorry, "program" the yaml definition you gave in codecov, do you mind giving it a kick to see how it feels?

nss/integration-tests/integration_test.go
```diff
@@ -153,23 +156,30 @@ func TestIntegration(t *testing.T) {
...
-		fmt.Fprintf(os.Stderr, "Can not build rust nss library: %v", err)
-		os.Exit(1)
-	}
-
 	testutils.InstallUpdateFlag()
 	flag.Parse()
 
-	m.Run()
+	code := m.Run()
+	defer os.Exit(code)
```
didrocks, line 163: Actually, as you noted, this is not needed:
As `MergeCoverages()` now returns an error and don’t `log.Fafal()`, the only part where we can "stop" the normal exit process is with your `defer os.Exit(24)`, which is garded by the `if code ==  0`, so we can safely remove this line.

nss/integration-tests/integration_test.go
didrocks: Just to check: the .profraw is in a subdirectory of the current repo? (I would prefer tempdir if possible, but otherwise, should be in `target/` so that we can avoid committing them by accident if something bad happens)

didrocks (changes requested):
Thanks for looking it up!

I think we can simplify one line and I have another question on the pprof. I really think we should avoid touching current directory

internal/testutils/coverage.go
```diff
@@ -0,0 +1,157 @@
...
+
+// MergeCoverages append all coverage files marked for merging to main Go Cover Profile.
+// This has to be called after m.Run() in TestMain so that the main go cover profile is created.
+// This has no action if profiling is not enabled.
+func MergeCoverages() error {
+	coveragesToMergeMu.Lock()
+	defer coveragesToMergeMu.Unlock()
+	for _, cov := range coveragesToMerge {
+		if err := appendToFile(cov, goMainCoverProfile); err != nil {
+			return fmt.Errorf("Teardown: can’t inject coverage into the golang one: %w", err)
```
didrocks, line 72: Go 1.20 bonus: I would use multiple errors here now that we can do it :)

didrocks: (TLDR: `err := errors.Join(err1, err2))`

nss/integration-tests/integration_test.go
denisonbarbosa: Right now they are being saved in the integration-tests directory (I'm not sure why)

internal/testutils/rust.go
```diff
@@ -118,8 +118,7 @@ func TrackRustCoverage(t *testing.T, src string) (env []string, target string) {
 	})
 
 	return []string{
-		"CARGO_INCREMENTAL=0",
-		"RUSTFLAGS=-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Cinstrument-coverage",
```
didrocks: hum, I thought you wanted to do the opposite, keeping the initial env variables that are advise by upstream to get more accurate profiling?

I just want to ensure this is done on purpose :)

internal/testutils/coverage.go
```diff
@@ -64,16 +65,20 @@ func TrackTestCoverage(t *testing.T) (testCoverFile string) {
...
 // This has no action if profiling is not enabled.
-func MergeCoverages() error {
+func MergeCoverages() (err error) {
 	coveragesToMergeMu.Lock()
 	defer coveragesToMergeMu.Unlock()
 	for _, cov := range coveragesToMerge {
-		if err := appendToFile(cov, goMainCoverProfile); err != nil {
-			return fmt.Errorf("Teardown: can’t inject coverage into the golang one: %w", err)
+		if tmpErr := appendToFile(cov, goMainCoverProfile); tmpErr != nil {
+			err = errors.Join(err, fmt.Errorf("%w", tmpErr))
```
didrocks, line 73: Thanks for changing it. However, it means that you need to bump go version in go.mod, ensure CI is all running with 1.20, bump the debian package dependency on 1.20

internal/testutils/coverage.go
```diff
@@ -64,16 +65,20 @@ func TrackTestCoverage(t *testing.T) (testCoverFile string) {
 // MergeCoverages append all coverage files marked for merging to main Go Cover Profile.
 // This has to be called after m.Run() in TestMain so that the main go cover profile is created.
 // This has no action if profiling is not enabled.
-func MergeCoverages() error {
+func MergeCoverages() (err error) {
```
didrocks, line 68: without decorate, as the return argument is obvious, I wouldn’t declare `err` here, but just `var err error` below. It’s just that implicit declarations are… hem (but necessary in case of decorate, unfortunately or to give hint to the caller about the input/output when necessary)

didrocks:
Some feedbacks, also, you could have seen that you should have bumped the Go version and both check quality and test CI are now failing.

I think we should have a Github option to not rerequest for a review until the CI pass, until then, it’s up to you to check that your PR pass CI :)

internal/testutils/rust.go
denisonbarbosa: It was done on purpose, indeed. These variables didn't seem to impact much in our coverage but test timing, but I can add them back and we see how it performs in the long time, wdyt?

internal/testutils/coverage.go
denisonbarbosa: Honestly, I wasn't expecting it to be approved hahaha I did this change mostly as a "joke" (because I thought your comment had the same intention). Do you feel like it's worth it to bump the go version just for this section?

Ps. I should've commented this yesterday, but I was so tired that I forgot. I'm sorry.

internal/testutils/coverage.go
denisonbarbosa: I'll fix it if we decide to go with this approach, but I doesn't really seem worth to bump the version just for it...

didrocks (approved):
LGTM now! Great to have coverage back! :)

PR #140, opened 2023-03-08
Updating auto-updates.yaml action

denisonbarbosa:
A few improvements on some of the rationales behind some chosen runners | actions;

Updating Rust related auto-updates to not merge the branch automatically.


.github/workflows/auto-updates.yaml
```diff
@@ -154,17 +170,21 @@ jobs:
...
           labels: control, automated pr
           body: "[Auto-generated pull request](https://github.com/ubuntu/adsys/actions/workflows/auto-updates.yaml) by GitHub Action"
           branch: auto-update-rust-checksum
+          delete-branch: true
           token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Push branch
-        if: ${{ steps.update-checksum.outputs.result == 'true' }}
-        run: |
-          git push origin auto-update-rust-checksum:main
+      - name: Automerge created Pull Request
```
didrocks, line 184: What does this gives us compared to only pushing the branch?

If I look at https://github.com/peter-evans/enable-pull-request-automerge#conditions, it seems to be really restrictive and moe complicated than a git push, I don’t see other simpification, hence my question.

denisonbarbosa: I'll resolve this since we already came to agreement on the right approach

.github/workflows/auto-updates.yaml
```diff
@@ -79,92 +84,47 @@ jobs:
...
           echo "modified=true" >> $GITHUB_ENV
         shell: bash
+      - name: Update the fake checksum file
+        run: |
+          set -eu
+
+          # Reads the checksum of both files
+          FAKE_CHECKSUM=$(cat debian/cargo-checksum.json | jq .package)
+          CRATE_CHECKSUM=$(cat vendor_rust/vcpkg/.cargo-checksum.json | jq .package)
+          if [ $FAKE_CHECKSUM == $CRATE_CHECKSUM ]; then
```
didrocks, line 101: Here. if FAKE_CHECKSUM or CREATE_CHECKSUM is empty for $reason, you will end up with a cryptic syntax error.

To read variables, it’s better to generally use that form: `${FAKE_CHECKSUM}`

.github/workflows/auto-updates.yaml
```diff
@@ -79,92 +84,47 @@ jobs:
...
+            echo "debian/cargo-checksum.json is up to date. No change is needed.";
+            exit 0
+          fi
+          # Formats in a way where we can replace the line using sed
+          CRATE_CHECKSUM="\"package\": $CRATE_CHECKSUM,"
```
didrocks, line 106: I would either use another variable, or do the string prefix directly in sed.

didrocks (changes requested):
Ensure you are testing it carefully in different scenarios, I have 2 comments on the script itself, otherwise, it’s way more concise and maintainable as you can see! :)

.github/workflows/auto-updates.yaml
denisonbarbosa: Ack, I will switch it

.github/workflows/auto-updates.yaml
denisonbarbosa: Yeah, I was doing it directly in `sed`, but I thought it would get confusing. I'll rename the variable to something like `FMT_CHECKSUM`, wdyt?

didrocks (approved):
LGTM!

PR #154, opened 2023-03-14
Replace testutils.Setenv

### LINKED ISSUE #117: Replace testutils.Setenv with t.SetEnv
Go has a t.SetEnv() from 1.17, remove our testutils function.

denisonbarbosa:
Since Go 1.17, the type `testing.T` has a method [t.Setenv](https://pkg.go.dev/testing#T.Setenv), which does precisely what we used to do in testutils.Setenv, so we should refactor the code to use the offered tools.

Closes #117 
DEENG-605


didrocks (approved):
easy! :)

PR #158, opened 2023-03-16
Replacing backend used by `rustix` crate

denisonbarbosa:
The rustix crate has the option of two backends to use: libc and linux_raw.
The linux_raw backend is shipped with .a files, which are the outlines that are statically linked at build time. Those files get removed when building the package and then "Houston, we have a problem". Usually this doesn't matter since the default way rustix builds itself is using asm, but this feature is experimental for ppc64el and MIPS architectures.
To fix this problem, we need to switch to using the libc backend, which is composed of only source files and, therefore, represents no issue to us.

DEENG-616

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -58,6 +58,7 @@ override_dh_auto_configure:
 	[ ! -e $(DH_CARGO_VENDORED_SOURCES) ] || $(DH_CARGO_VENDORED_SOURCES)
 	DEB_HOST_GNU_TYPE=$(DEB_HOST_GNU_TYPE) DEB_HOST_RUST_TYPE=$(DEB_HOST_RUST_TYPE) \
 		CARGO_HOME=$(CURDIR)/debian/cargo_home DEB_CARGO_CRATE=nss_aad_$(shell dpkg-parsechangelog --show-field Version) \
+		RUSTFLAGS="--cfg=rustix_use_libc" \
 		$(CARGO) prepare-debian $(CARGO_VENDOR_DIR)
 
 	# By default, dh-cargo only replaces the crates-io source, which means we have to replace the git source manually


didrocks (commented):
One question before approving: can you check the RUSTFLAGS env variable (if dh_cargo set it to something?), so that we don’t override distro's default (and we need to append that flag to the default ones?)

denisonbarbosa:
> One question before approving: can you check the RUSTFLAGS env variable (if dh_cargo set it to something?), so that we don’t override distro's default (and we need to append that flag to the default ones?)

The Rust doc states that `RUSTFLAGS` is a "list of custom flags to pass to all compiler invocations that cargo performs" and it has no default value as well. Ideally, we would use it just when building the lib for the tests, but since the rustix crate is only used during tests there's no difference.
About what dh_cargo does: it reads the content of the env and if there's some, it adds those flags to a [build] section on the config.toml file that it generates while also unsetting the `RUSTFLAGS`, so we are safe.

didrocks:
> About what dh_cargo does: it reads the content of the env and if there's some, it adds those flags to a [build] section on the config.toml file that it generates while also unsetting the RUSTFLAGS, so we are safe.

Yes, I was after that behaviour, but we don’t use dh_cargo directly but the wrapper called by dh_cargo. So I think we should look at what dh_cargo does itself before calling the wrapper in term of environment variables swapping/append and do the same if it does any.

denisonbarbosa:
> > About what dh_cargo does: it reads the content of the env and if there's some, it adds those flags to a [build] section on the config.toml file that it generates while also unsetting the RUSTFLAGS, so we are safe.
> 
> Yes, I was after that behaviour, but we don’t use dh_cargo directly but the wrapper called by dh_cargo. So I think we should look at what dh_cargo does itself before calling the wrapper in term of environment variables swapping/append and do the same if it does any.

Alright, so I've looked into it and dh_cargo does not use the env RUSTFLAGS, so we are not overriding anything.

didrocks (approved):
Thanks for checking!

PR #184, opened 2023-04-17
Switching libnss dependency source branch

### LINKED ISSUE #179: Investigate CI tests failing for NSS Group queries
CI tests are failing with `Expected no error but got one: signal: aborted (core dumped)` for Group queries on `NSS`. The debug message shows `thread '<unnamed>' panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x55bc0ac5fca7', /home/runner/.cargo/git/checkouts/libnss-rs-4e6320ade2d0483d/02e0fa4/libnss/src/interop.rs:161:13`.

This seems to be a problem with the pointers of Group, since it has a Vec in its field.

denisonbarbosa:
Now we have 2 fixes to submit to upstream. Since they have to be on different PRs, we had to create a branch that would glob all commits for us to use in aad-auth.

Closes #179 / DEENG-651

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,7 +264,7 @@ checksum = "99227334921fae1a979cf0bfdfcc6b3e5ce376ef57e16fb6fb3ea2ed6095f80c"
 [[package]]
 name = "libnss"
 version = "0.4.0"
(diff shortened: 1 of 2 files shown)


didrocks (approved):
Excellent, thanks for the quick investigation!

PR #193, opened 2023-04-22
Improve error messages for Azure AD error codes

peat-psuwit:
This PR improve error messages for 2 cases:

- Provides useful diagnostics for Azure AD code `65001` and `7000218` [1], which can be found when the Azure AD application is mis-configured. Error `65001` could even be resolved by just following the link in the error message.
- Provides links, similar to [1], for other unknown codes. It links to Microsoft's error code website, which should helps when debugging the issue.

[1]
- https://login.microsoftonline.com/error?code=65001
- https://login.microsoftonline.com/error?code=7000218


didrocks:
Hey! Thanks for this enhancements and making aad-auth better. The code itself looks good and will help a lot of users as we saw bug reports around this.

There are two things to consider before merging:
- can you sign the CLA (you have a link in the bot description) and tell us once done?
- this project is heavily tested. Consequently, those changes should test every error cases handling you added. The corresponding test is in https://github.com/ubuntu/aad-auth/blob/main/internal/aad/aad_test.go#L13 (you need to add 2 cases). This will trigger the mock to behave differently and returning different errors based on the user name. You can see some example of this in https://github.com/ubuntu/aad-auth/blob/main/internal/aad/msal_mock.go#L47.

Do not hesitate if this triggers any questions or if you need any help! Thanks again :)

peat-psuwit:
I've actually signed CLA right after I submitted this PR, but I guess it has to be checked by a human before an entitlement is added. Anyway, should be solved now.

The test cases are added.

internal/aad/aad_test.go
```diff
@@ -28,6 +28,8 @@ func TestAuthenticate(t *testing.T) {
 		"invalid server response":    {username: "invalid server response", wantErr: aad.ErrDeny},
 		"invalid credentials":        {username: "invalid credentials", wantErr: aad.ErrDeny},
 		"no such user":               {username: "no such user", wantErr: aad.ErrDeny},
+		"public client disallowed":   {appID: "public client disallowed", wantErr: aad.ErrDeny},
+		"no tenant-wide consent":     {appID: "no tenant-wide consent", wantErr: aad.ErrDeny},
```
denisonbarbosa, line 32: Would you mind adding those right after the `"can't connect to authority"` case? Since those error cases are related to external errors, it's better to have them grouped together.

internal/aad/aad.go
```diff
@@ -86,8 +88,25 @@ func (auth AAD) Authenticate(ctx context.Context, cfg config.AAD, username, pass
...
+				logger.Err(ctx, "Azure AD application requires enabling 'Allow public client flows'. "+
+					"https://learn.microsoft.com/en-us/azure/active-directory/develop/scenario-desktop-app-registration#redirect-uris")
+				return ErrDeny
+			}
 		}
 		logger.Err(ctx, "Unknown error code(s) from server: %v", addErrWithCodes.ErrorCodes)
+
+		logger.Debug(ctx, "For more information about the error code(s), see:")
+		for _, errcode := range addErrWithCodes.ErrorCodes {
+			logger.Debug(ctx, "- https://login.microsoftonline.com/error?code=%d", errcode)
```
denisonbarbosa, line 107: Can you improve the debug log with the error code specified out of the URL as well?
i.e.:
`"Error code: %d  - https://login.microsoftonline.com/error?code=%d"`

Even though this creates a bit of redundancy, it improves the readability and makes it easier for the user to spot the URL that relates to the code that matters.

denisonbarbosa (commented):
LGTM. Thanks for the changes!
A few minor nitpicks to improve readability, but everything else is great.

internal/aad/aad.go
peat-psuwit: How about the one I updated?

denisonbarbosa (approved):
Looks great now! Thanks for addressing the suggestions.

internal/aad/aad.go
denisonbarbosa: Looks good as well, thanks!

didrocks (approved):
Excellent work! Thanks for your help here, I’m sure it will unlock a lot of users who wants to set that up.

PR #215, opened 2023-05-02
Shadow mode eval now considers effective ids

### LINKED ISSUE #197: Unable to `sudo` anything while logged in as AAD user
Description / Steps to Re-produce:

1) Login in using an AAD account (I'm a global admin)
2) Open Terminal Windows
3) Try to use `sudo`, for example `sudo pro attach [key]`
4) Type in account password when requested

Always returns: Sorry, try again

My password is correct (it does have Upper, Lower and Symbols in it)

My local account (which has the same password), works if I switch users etc.

denisonbarbosa:
In order to determine the shadow mode of the cache, we were using `unix.Access()`, which considers only the real uid and gid of the current process. This was causing `sudo` calls to fail, since they only change the `effective` uid of the current process.

Replacing `unix.Access()` by `unix.Faccessat()` allows us to consider the `effective` uid of the caller instead, allowing `sudo` calls to work if the user is in the `sudo` group or in the `/etc/sudoers` file.

Closes #197 
DEENG-664

### DIFF
--- a/internal/cache/cachedb.go
+++ b/internal/cache/cachedb.go
@@ -113,11 +113,12 @@ func initDB(ctx context.Context, cacheDir string, rootUID, rootGID, shadowGID, f
 	// Attach shadow if our user has access to the file (even read-only)
 	shadowMode = forceShadowMode
 	if forceShadowMode == -1 {
-		shadowMode = 0
-		if unix.Access(shadowPath, unix.W_OK) == nil {
-			shadowMode = shadowRWMode
-		} else if unix.Access(shadowPath, unix.R_OK) == nil {
+		shadowMode = shadowNotAvailableMode
+		if unix.Faccessat(unix.AT_FDCWD, shadowPath, unix.R_OK, unix.AT_EACCESS) == nil {
 			shadowMode = shadowROMode
+		}
+		if unix.Faccessat(unix.AT_FDCWD, shadowPath, unix.W_OK, unix.AT_EACCESS) == nil {
+			shadowMode = shadowRWMode
 		}
 	}
 	if shadowMode > shadowNotAvailableMode {


johnkattenhorn:
Great - work - can you let us know when this will be available for us to pull as an update ?

denisonbarbosa:
Hey @johnkattenhorn, it might take a while before we get the security update released. I can upload it to a PPA for you to test it, but I don't recommend using it in production yet.

Also, the correct way of giving sudo privileges to a user is to add it to the sudo group, but since this is a general concept rather than a specificity of `aad-auth`, it doesn't belong in its documentation.

johnkattenhorn:
OK - do we already have something in the documentation about Security Groups from Azure - my only feedback from our team  (who are Windows admins mostly) was they were expecting that some magic was supposed to happen between Global Admins and sudo ;-) Maybe we should call out that standard sudo rules apply here. 

The PPA would be super useful

denisonbarbosa:
Here is the ppa for the fixed version: https://launchpad.net/~justdenis/+archive/ubuntu/aad-auth.

velayudhamv:
When will this be available for general to test? Will there be any update related to this in the repo?

John-Vance:
@denisonbarbosa Just an FYI, I used your PPA for 0.4.1 and it seems to have fixed everything perfectly. Thanks so much!

velayudhamv:
Is this correct?
sudo add-apt-repository ppa:justdenis/test-aad-auth
sudo apt update

sudo apt install libpam-aad libnss-aad

Or any other command?

John-Vance:
@velayudhamv That's exactly what I did, worked for me!

kirkog86:
BTW, aad-auth will work on 20.04/22.04, using the below :)

sudo add-apt-repository ppa:justdenis/test-aad-auth
sed -i "s/$(lsb_release -c | awk '{print $2}')/lunar/g" /etc/apt/sources.list.d/justdenis-ubuntu-test-aad-auth-$(lsb_release -c | awk '{print $2}').list
sudo apt update
sudo apt install libpam-aad libnss-aad

igovea-ea:
> Here is the ppa for the fixed version: https://launchpad.net/~justdenis/+archive/ubuntu/test-aad-auth.

This is no longer accessible for me.

erdelmaero:
> > Here is the ppa for the fixed version: https://launchpad.net/~justdenis/+archive/ubuntu/test-aad-auth.
> 
> This is no longer accessible for me, attempting to add the ppa via Ubuntu 23.04 notes that it does not exist either.

The ppa seems to be gone. It fixed everything for me...

denisonbarbosa:
Hey guys, the fixed version is now available in [ppa:justdenis/aad-sudo-fix](https://launchpad.net/~justdenis/+archive/ubuntu/aad-sudo-fix)

PR #279, opened 2023-06-29
Add resolver definition to root manifest

denisonbarbosa:
Our CI started failing when building the `proc-macro2` crate and cargo gave us some notes:
```text
Setup: could not build Rust NSS library: warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify
`workspace.resolver = "2"` in the workspace root's manifest
```

So we should lock the resolver to version "2" directly on the root Cargo manifest and turn the `proc-macro2` transitive dependency into an actual one, so we can better control its version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -337,6 +337,7 @@ dependencies = [
  "libnss",
  "log",
  "paste 1.0.12",
+ "proc-macro2",
  "rusqlite",
  "serde",
  "serde_yaml",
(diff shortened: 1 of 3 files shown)


didrocks:
Any reason we can’t switch to version 2 directly? Seems like the way to move forward if everything is compatible, no?

denisonbarbosa:
> Any reason we can’t switch to version 2 directly? Seems like the way to move forward if everything is compatible, no?

Well, that's a good point :joy: I've tried it and it still works, so let's go forward (as it's definitely the best way to go)!

didrocks (approved):
way better, indeed!

PR #327, opened 2023-08-18
Fix FTBFS due to missing new required symbol

denisonbarbosa:
Golang 1.21 introduced a new required symbol `x_crosscall2_ptr` so we need to update the `libpam-aad.symbols` file to prevent FTBFS.


didrocks:
As the new symbol is related to Go 1.21, this should become the minimal build requirement in debian/control.

PR #335, opened 2023-08-24
Release 0.5.2

denisonbarbosa:
The fixes that we submitted upstream for the `libnss` crate got merged and released in version `0.5` of the crate. Since we also had an MSAL update (to `1.1.1`), we should do a minor new release of aad-auth with the updates and clean up some of the packaging steps that were needed to use our fork repo with the fixes.

PPA with the builds: https://launchpad.net/~justdenis/+archive/ubuntu/mantic-aad-proposed/

UDENG-1245


debian/changelog
```diff
@@ -1,3 +1,20 @@
+aad-auth (0.5.2) mantic; urgency=medium
+
+  * Update Go version to 1.21
```
didrocks, line 3: I would be more precise the the update is in debian/control for go build dependency telling that the symbol file were already requiring them.

didrocks (changes requested):
one slight comment change!

debian/changelog
denisonbarbosa: So we shouldn't touch the required version in `go.mod`?

didrocks: we do, it should have been part of 0.5.1 upload anyway (actually both, debian/control and this) as the symbol file depends on it. Forget about the debian/control explicitness, but tell in the changelog that this is not a new feature-breakage, but basically something we have already done in the last build.

didrocks (approved):

PR #371, opened 2023-10-22
Passwddb locking fix (nss_aad: database error: database is locked)

johne8:
fix: nss_aad: database error: database is locked

ref: aad_auth[2196]: nss_aad: database error: database is locked

This adresses this, by using Write-ahead logging journaling mode when openening the database in ReadWrite mode, 
and opening the database immutable when opening it in ReadOnly mode.

### DIFF
--- a/internal/cache/cachedb.go
+++ b/internal/cache/cachedb.go
@@ -3,6 +3,7 @@ package cache
 import (
 	"context"
 	"database/sql"
+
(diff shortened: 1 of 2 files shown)


johne8:
Will squash commits,
 also forgot to add PRAGMA journal_mode=wal; to the passwd.sql file

PR #376, opened 2023-10-25
Passwddb locking fix (nss_aad: database error: database is locked)

johne8:
fix: nss_aad: database error: database is locked

ref: aad_auth[2196]: nss_aad: database error: database is locked

This adresses this, by using Write-ahead logging journaling mode when openening the database in ReadWrite mode,
and opening the database immutable when opening it in ReadOnly mode.

PASS when build and test with autopkgtest.


internal/cache/cachedb.go
```diff
@@ -25,6 +25,7 @@ const (
 	defaultCachePath = "/var/lib/aad/cache"
 	passwdDB         = "passwd.db" // root:root 644
 	shadowDB         = "shadow.db" // root:shadow 640
+	db_con_args      = "?_journal_mode=wal"
```
denisonbarbosa, line 28: Go does not allow snake casing in variable names, would you mind refactoring this?
```suggestion
	dbConnArgs      = "?_journal_mode=wal"
```

nss/src/cache/mod.rs
```diff
@@ -300,7 +304,15 @@ impl CacheDBBuilder {
 
         let passwd_db = &db_path.join(PASSWD_DB);
         let passwd_db = passwd_db.to_str().unwrap();
-        let conn = match Connection::open_with_flags(passwd_db, open_flags) {
+    
+        let mut passwd_db_args = DB_CON_OPT_RO;
+        if shadow_mode == ShadowMode::ReadWrite { 
+            passwd_db_args = DB_CON_OPT_RW;
+        }
+        let passwd_db_con = format!("{}{}{}", DB_CON_PREFIX.to_string(), passwd_db, passwd_db_args);
```
denisonbarbosa, line 312: There's no need to call `.to_string()` in a &str type when formatting or printing it.

nss/src/cache/mod.rs
```diff
@@ -309,6 +321,7 @@ impl CacheDBBuilder {
         if shadow_mode >= ShadowMode::ReadOnly {
             let shadow_db = shadow_db.to_str().unwrap();
 
+            debug!("attaching shadow database");
```
denisonbarbosa, line 324: It's not a good idea to have this type of information printed on the system logs. Would you mind removing this?

cmd/aad-cli/cli/testdata/cachedb/passwd.db
denisonbarbosa, line 1: Would you mind taking another look here? Looking at the PR changes, I don't see why there's a change in the database file. Maybe a change was made during manual testing?

nss/src/cache/mod.rs
```diff
@@ -26,6 +26,10 @@ pub const PASSWD_PERMS: u32 = 0o644;
 const SHADOW_DB: &str = "shadow.db"; // Ownership: root:shadow
 pub const SHADOW_PERMS: u32 = 0o640;
 
+const DB_CON_PREFIX: &str = "file:"; // Specify "file" as access mode, so that we can use connection options.
+const DB_CON_OPT_RW: &str = "?journal_mode=wal"; // Use Write Ahead Log journaling mode, so that we can operate paralell with the PAM module.
+const DB_CON_OPT_RO: &str = "?immutable=1"; // When using immutable=1, we can still read the passwd_db, even if there is an db lock.
```
denisonbarbosa, line 31: Minor nitpick here: I would refactor these names to use `CONN` instead of `CON`, as it gives a better idea of what this value refers to.

denisonbarbosa (changes requested):
Hey @johne8, thanks for helping with this! I like the changes, but there are a few minor issues (and some nitpicks) that I'd like to see fixed. 
Apart from those, all looks good. Well done!

internal/cache/cachedb.go
johne8: Hello, thanks for the feedback. fixed in latest commit. :)

nss/src/cache/mod.rs
johne8: Hello, thanks for the feedback. fixed in latest commit. :)

nss/src/cache/mod.rs
johne8: It's removed in the latest commit, I used this for debugging purposes.

cmd/aad-cli/cli/testdata/cachedb/passwd.db
johne8: The file is changed, I added "PRAGMA journal_mode=wal;" (it does not print if you use ".dump"), the original test database file had journal_mode=DELETE;

nss/src/cache/mod.rs
johne8: Thanks! good catch, its chaneged in the latest commit.

cmd/aad-cli/cli/testdata/cachedb/passwd.db
johne8: you can verify with: 
sqlite3 cmd/aad-cli/cli/testdata/cachedb/passwd.db 'PRAGMA journal_mode;'
It should return: 'wal'

internal/cache/cachedb.go
```diff
@@ -25,6 +25,7 @@ const (
 	defaultCachePath = "/var/lib/aad/cache"
 	passwdDB         = "passwd.db" // root:root 644
 	shadowDB         = "shadow.db" // root:shadow 640
+	dbConnArgs      = "?_journal_mode=wal"
```
denisonbarbosa, line 28: After the change, the `=` became misaligned. Can you fix it to calm down the linter?

johne8: Yes, thanks for catching that. :)
fixed in latest commit, I also updated the branch.

denisonbarbosa (commented):
Thanks for addressing the comments, @johne8! 
I'll do a round of manual testing before approving the pull request. Meanwhile, can you rewrite the commit history so we can have a clean history about what matters?

denisonbarbosa (approved):
LGTM. Thanks a lot for tackling this, @johne8! We appreciate the help and contribution. Well done!

Issue #399, opened 2023-11-14
Issue: podman with "--userns=keep-id" failes. 


johne8:
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

Seems like the permissions for the filesystem get spoofed when using podman with "--userns=keep-id"

in the logfile i get
  aad_auth[7954]: nss_aad: database error: invalid ownership for /var/lib/aad/cache/passwd.db, expected 0:0 but got 65534:65534
  where "7954" equals the PID of podman (non-root)
  
I can fix this by doing the following changes to nss/src/cache/mod.rs (line 369):

```
            // skip ownership check if detected owned by nobody.
            if stat.uid() != 65534 {
                // Checks ownership
                if stat.uid() != file.expected_uid || stat.gid() != file.expected_gid {
                    return Err(CacheError::DatabaseError(format!(
                        "invalid ownership for {}, expected {}:{} but got {}:{}",
                        file.path.to_str().unwrap(),
                        file.expected_uid,
                        file.expected_gid,
                        stat.uid(),
                        stat.gid()
                    )));
                }
            }
```

### Steps to reproduce it

* make sure you have subuid and subgid added for the entraid user in /etc/subuid and /etc/subgid.
   example:   user@tenant.onmicrosoft.com:165536:65536
* login with the entraid user.
* run example: 
     podman run --rm -it --userns=keep-id ubuntu:latest bash
you will get error: 
`Error: error creating temporary passwd file for container  {CONTAINERID}: failed to get current user: user: unknown userid {USERID}`

### Ubuntu users: System information and logs

journalctl -f | grep "_aad"
 aad_auth[7954]: nss_aad: database error: invalid ownership for /var/lib/aad/cache/passwd.db, expected 0:0 but got 65534:65534

### Non Ubuntu users: System information and logs

im using ubuntu

### Double check your logs

- [X] I have redacted any sensitive information from the logs

johne8:
would you like me to add a pull request for my "workaround"?

PR #400, opened 2023-11-14
Issue: #399 Update mod.rs, quickfix workaround

johne8:
Workaround for issue #399
allow podman to run with --userns=keep-id

### DIFF
--- a/nss/src/cache/mod.rs
+++ b/nss/src/cache/mod.rs
@@ -366,16 +366,19 @@ impl CacheDBBuilder {
                 )));
             }
 
-            // Checks ownership
-            if stat.uid() != file.expected_uid || stat.gid() != file.expected_gid {
-                return Err(CacheError::DatabaseError(format!(
-                    "invalid ownership for {}, expected {}:{} but got {}:{}",
-                    file.path.to_str().unwrap(),
-                    file.expected_uid,
-                    file.expected_gid,
-                    stat.uid(),
-                    stat.gid()
-                )));
+            // skip ownership check if detected owned by nobody.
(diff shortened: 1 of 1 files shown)


didrocks (commented):
@johne8: thanks for looking at this and help at figuring out your issue with podman!

I don’t really like that we hardcode the ownership integer value of nobody, which can vary depending on the distribution and user config.

Secondly, this check is to ensure an integrity of the database and poking a hole there doesn’t seem like the right solution. I think looking at the user namespace mapping might help there to ensure that in the parent namespace, the uid/gid still match the expecations, wdyt?

johne8:
> @johne8: thanks for looking at this and help at figuring out your issue with podman!
> 
> I don’t really like that we hardcode the ownership integer value of nobody, which can vary depending on the distribution and user config.
> 
> Secondly, this check is to ensure an integrity of the database and poking a hole there doesn’t seem like the right solution. I think looking at the user namespace mapping might help there to ensure that in the parent namespace, the uid/gid still match the expecations, wdyt?

hmm, I'm not too familiar with namespaces. The best thing I can think of currently is reading the values of: /proc/sys/kernel/overflow**u**id and /proc/sys/kernel/overflow**g**id, then check if the uid and gid owner matches that.

johne8:
Have you checked out my latest commit? It seems to work fine :) 

Also in Ubuntu 22.04 LTS, it seems like UID/GID above 2147483647 (SignedInt) causes issues in gnome.

So needed to add some workaround for this also by reducing the UID of the users that have a UID calculated above that value.

PR #413, opened 2023-12-12
deps: bump go version

jibel:
Addresses vulnerability fixes.


didrocks:
As a side note, that should  be the toolchain to get updated, not the go line. I think it’s fine for aad-auth though.

PR #434, opened 2024-01-05
ci: update deps weekly

jibel:
Reduce the noise by doing dependency check weekly.


denisonbarbosa (approved):
Looks good! I think it's better this way as well.
Can you squash the changes to a single commit before merging?

PR #451, opened 2024-02-07
deps: Update deps to latest available SemVer version

denisonbarbosa:
This is the result of running `cargo update`. Dependabot does not catch dependency updates for transitive dependencies in Rust, so we should also consider adding this command to a CI job.


jibel (approved):
+1

Thank you Sir!

